PR #50, opened 2018-02-02
Change type on composer.json

mistermarco:
With composer/installers and type: drupal-custom-theme, this theme would automatically be installed in /modules/custom when required by a Drupal project. The original stanford-theme type wasn't recognized and so the theme was installed in the /vendor directory.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,10 +1,12 @@
 {
   "name": "su-sws/stanford_basic",
   "description": "Stanford Basic Branding Theme.",
-  "type": "stanford-theme",
+  "type": "drupal-custom-theme",
   "homepage": "https://github.com/SU-SWS/stanford_basic",
   "authors": [],
   "license": "GPL-2.0+",
   "minimum-stability": "dev",
-  "require": {}
+  "require": {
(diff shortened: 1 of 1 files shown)


sherakama:
In our composer projects we have the following snippet in order to separate out Stanford shared modules/themes from project custom modules and themes.
```
...
"custom-installer": {
      "docroot/modules/stanford/{$name}/": [
        "type:stanford-module"
      ],
      "docroot/themes/stanford/{$name}/": [
        "type:stanford-theme"
      ],
...
```
I can see how the default setting of drupal-custom-theme would be nice for others using the project. Going to have to ponder the pros/cons of this for us. Without this keyword we would have to individually set the path for each project if we wanted to separate them in to a Stanford folder.

mistermarco:
Aha! I knew there had to be a reason you did it that way. Understood. Thanks!

sherakama:
Im going to re-open this ticket. It is a good discussion point. I'll bring it up with the other devs.

sherakama:
We have updated this in newer branches but have decided to leave it as is on the 2.x branch.

PR #51, opened 2018-02-21
MRCD8-45 added apple touch icons

josephgknox:


sherakama:
If you can provide the path to the icons why not put these in the img folder?

josephgknox:
@sherakama thanks, Shea! I'll move these to an img folder, only leaving the favicon as the root.

Cc: @pookmish

PR #59, opened 2018-04-18
added core library dependencies

pookmish:
# READY FOR REVIEW

# Summary
- Add normalize and modernize as dependencies 
- This fixes alot of margins and gaps.
- Adds styling for messages so they aren't just plain text.

# Needed By (Date)
- whenever

# Urgency
- low

# Steps to Test

1. spin up new site
2. add the theme and enable
3. ensure the branding bar at the top doesn't have any whitespace around it.

# See Also


stanford_basic.libraries.yml
```diff
@@ -102,6 +102,13 @@ basic:
       css/theme/print.css: { media: print }
   js:
     js/vendor/modernizr.js: {}
+  dependencies:
+    - classy/base
+    - classy/messages
+    - classy/progress
+    - core/modernizr
+    - core/normalize
```
sherakama: Normalize is in decanter which gets rendered in to the css/base/base.css stylesheet and thus is not a dependency here.

stanford_basic.libraries.yml
```diff
@@ -102,6 +102,13 @@ basic:
       css/theme/print.css: { media: print }
   js:
     js/vendor/modernizr.js: {}
+  dependencies:
+    - classy/base
+    - classy/messages
+    - classy/progress
+    - core/modernizr
```
sherakama: This dupes the script at js/vendor/modernizr.js please pick one.

stanford_basic.libraries.yml
```diff
@@ -102,6 +102,13 @@ basic:
...
+    - core/normalize
+    - core/jquery
```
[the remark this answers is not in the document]
pookmish: maybe not right now. maybe later we will. could remove it until then

stanford_basic.libraries.yml
pookmish: I didnt see the normalize taking affect without defining the dependency

stanford_basic.libraries.yml
pookmish: fair nuff

stanford_basic.libraries.yml
```diff
@@ -102,6 +102,13 @@ basic:
       css/theme/print.css: { media: print }
   js:
     js/vendor/modernizr.js: {}
+  dependencies:
+    - classy/base
```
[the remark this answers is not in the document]
pookmish: It adds styles for forms, pagers, and a ton of extras:
```      css/components/action-links.css: { weight: -10 }
      css/components/breadcrumb.css: { weight: -10 }
      css/components/button.css: { weight: -10 }
      css/components/collapse-processed.css: { weight: -10 }
      css/components/container-inline.css: { weight: -10 }
      css/components/details.css: { weight: -10 }
      css/components/exposed-filters.css: { weight: -10 }
      css/components/field.css: { weight: -10 }
      css/components/form.css: { weight: -10 }
      css/components/icons.css: { weight: -10 }
      css/components/inline-form.css: { weight: -10 }
      css/components/item-list.css: { weight: -10 }
      css/components/link.css: { weight: -10 }
      css/components/links.css: { weight: -10 }
      css/components/menu.css: { weight: -10 }
      css/components/more-link.css: { weight: -10 }
      css/components/pager.css: { weight: -10 }
      css/components/tabledrag.css: { weight: -10 }
      css/components/tableselect.css: { weight: -10 }
      css/components/tablesort.css: { weight: -10 }
      css/components/tabs.css: { weight: -10 }
      css/components/textarea.css: { weight: -10 }
      css/components/ui-dialog.css: { weight: -10 }```

File: stanford_basic.libraries.yml (hunk 3 above)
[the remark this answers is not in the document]
Comment (sherakama) [reply stanford_basic.libraries.yml]: @boznik @josephgknox @minorwm 

How do you feel about adding jquery to every page? Would this help or hurt your work? jquery was always available in D7 and can make sense for this theme.

File: stanford_basic.libraries.yml (hunk 1 above)
Comment (sherakama) [reply stanford_basic.libraries.yml]: Sounds like a bug or perhaps a load order issue. Some more investigation is needed.

Comment (pookmish) [reply stanford_basic.libraries.yml]: Heres a screenshot of commenting out the normalize and also using normalize
screen shot 2018-04-18 at 9 42 36 am
screen shot 2018-04-18 at 9 43 09 am

File: stanford_basic.libraries.yml (hunk 4 above)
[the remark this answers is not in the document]
Comment (sherakama) [reply stanford_basic.libraries.yml]: @josephgknox 

How does this sync up with your work? Looks like some overlap between the efforts in decanter/basic.

File: stanford_basic.libraries.yml (hunk 1 above)
Comment (pookmish) [reply stanford_basic.libraries.yml]: maybe something from decanter isn't populating?

File: stanford_basic.libraries.yml (hunk 4 above)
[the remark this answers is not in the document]
Comment (josephgknox) [reply stanford_basic.libraries.yml]: @sherakama @pookmish upon looking through the above list quickly, none of them conflict in any way that concerns me immediately, with the exception of buttons and links. Those two are being defined and styled in Decanter. They do have related classes in Decanter that would likely override the default presented in Classy, but am not sure... other than those two, though, I;m cool with leaving them. No sense in reinventing the wheel with those basic items.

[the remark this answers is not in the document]
Comment (pookmish) [reply stanford_basic.libraries.yml]: For MRC it was a quick fix to add the astricks for a required field.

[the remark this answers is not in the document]
Comment (josephgknox) [reply stanford_basic.libraries.yml]: +1 -- things like that proved useful for MRC and likely will on all sites. Is it more beneficial to cherrypick what we want and add to Decanter or to add Classy as a dependency?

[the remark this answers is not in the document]
Comment (sherakama) [reply stanford_basic.libraries.yml]: Cool, just checkin in on an organization standpoint. Makes sense to pull those in to basic if they are available and useful. Just want to be sure that some of those base(er) components do not overlap or get lost from the work in the base styles of Decanter. Thanks for the feedback.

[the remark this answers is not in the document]
Comment (josephgknox) [reply stanford_basic.libraries.yml]: For sure! Thanks for always keeping us honest. 👊

File: stanford_basic.libraries.yml (hunk 2 above)
Comment (sherakama) [reply stanford_basic.libraries.yml]: @pookmish, It might be nice to rely on the core version instead of our own. Are they the same version? Could we switch to core?

Comment (pookmish) [reply stanford_basic.libraries.yml]: Core is actually using version 3.3.1.
our theme is currently on 3.2.0

File: stanford_basic.libraries.yml (hunk 1 above)
Comment (sherakama) [reply stanford_basic.libraries.yml]: Normalize is being rendered in to the base.css at: https://github.com/SU-SWS/stanford_basic/blob/8.x-2.x/css/base/base.css#L2337

It is after font-awesome. So perhaps something is up with that?

Also, look at where base.css is in the order of being added to the page (the cascade) and see if it is too far after other stylesheets. This could be causing issues as well. Might have to put a  light weight on it.

File: stanford_basic.libraries.yml (hunk 2 above)
Comment (sherakama) [reply stanford_basic.libraries.yml]: AFAIK we don't have any scripts that rely on modernizr. Let's switch to the core dependency.

Comment (pookmish) [reply stanford_basic.libraries.yml]: I deleted the local modernizr file

File: stanford_basic.libraries.yml (hunk 1 above)
Comment (pookmish) [reply stanford_basic.libraries.yml]: Core uses the same library, but the one in core contains slightly different css. perhaps if we update ours?

Comment (pookmish) [reply stanford_basic.libraries.yml]:
```normalize:
  remote: https://github.com/necolas/normalize.css
  version: "3.0.3"
  license:
    name: MIT
    url: https://github.com/necolas/normalize.css/blob/master/LICENSE.md
    gpl-compatible: true
  css:
    base:
      assets/vendor/normalize-css/normalize.css: { weight: -20 }
```

pookmish: just noticed the version differences. that could be why. core: 3.0.3 us: 6.0.0

sherakama: @josephgknox @pookmish 

I'd like normalize to still be a dependency of Decanter as it is needed for that project itself but I see where this could conflict with Drupal's. Normalize is not the only thing that could conflict as Drupal does have a number of core libraries now. 

@josephgknox perhaps we need a way to include decanter without FA or Normalize? That might be too much of a difficult task so I would suggest punting and lean on the code that comes out of Decanter's render at this point.

Patch up what you need in stanford_basic's base.css output after including decanter.scss

stanford_basic.libraries.yml
sherakama: Great. Thank you.

josephgknox:
Sounds good to me.

stanford_basic.libraries.yml
```diff
@@ -100,8 +100,12 @@ basic:
     theme:
       css/theme/theme.css: {}
       css/theme/print.css: { media: print }
-  js:
-    js/vendor/modernizr.js: {}
+  dependencies:
+    - classy/base
+    - classy/messages
+    - classy/progress
+    - core/normalize
```
sherakama: Please remove this as normalize is in decanter and gets rendered to base.css

pookmish: Are we thinking about changing the normalize in decanter? since there seems to be some items missing for drupal

sherakama: @josephgknox can you work with @pookmish to determine what is missing for this theme in the normalize package that is in Decanter? I'd prefer to stay at the later versions but if you have a convincing argument as to why Decanter should be on a previous version I could be swayed. Would it be possible to 'patch up' the missing items in stanford_basic instead?

Also, it looks like normalize is on version 8 now. Perhaps we upgrade?

Current version in Decanter: 6.0.0
Current version in Drupal: 3.0.3

Diff [3.0.3 to 6.0.0](https://github.com/necolas/normalize.css/compare/3.0.3...6.0.0)
Diff [6.0.0 to 8.0.0](https://github.com/necolas/normalize.css/compare/6.0.0...8.0.0)
Diff [3.0.0 to 8.0.0](https://github.com/necolas/normalize.css/compare/3.0.3...8.0.0)

pookmish: Hmm id be interested to see if it fixes by upgrading. but i think patching in stanford_basic would be a good aproach

josephgknox: Chatted with Mike. Only way to be sure is to test. I'll update to 8.0 in Decanter and we can recompile in Stanford Basic (on this branch).

josephgknox: https://github.com/SU-SWS/decanter/pull/101

josephgknox:
@sherakama @pookmish normalize has been updated in Decanter and stanford_basic. Also, all of the the shared dependencies (that is, dependencies that are called in both projects) are now on the same version.

pookmish:
Updating normalize in decanter seems to have fixed things within drupal too. currently nothing to over-ride in this theme. So to summarize:
added these 3 libraries:
```    
    - classy/base
    - classy/messages
    - classy/progress
```
and removed modernizr in favor of drupal core version.

PR #62, opened 2018-05-02
Fa fixup

josephgknox:
# READY FOR REVIEW

# Summary
- Update to latest Decanter and compile
- Change how FontAwesome is being added to project to help with issues being inherited down the line (Decanter will no longer have FA and neither with sub-themes. Rather, stanford_basic will host FA for use down the chain)

# Needed By (Date)
- N/A

# Urgency
- N/A

# Steps to Test

1. Pull this branch
2. Ensure that latest Decanter has been added (a good test: new normalize is present and FontAwesome is gone)
3. Verify that FontAwesome is being added to the project as needed

- stanford_basic
- su_humsci_theme
- decanter

- N/A

# See Also


libraries/decanter/templates/elements/links/links.html.twig
```diff
@@ -0,0 +1 @@
+<p>This is a <a class="{{ modifier_class }}" href="#">{{ linkType [0] }}</a></p>
```
pookmish: We might need to namespace these template files in decanter. Drupal is picking it up and using it as the link template.
cc @sherakama

sherakama: huh, I thought Drupal only looked for these in the /templates directory.

I wonder if you can put a namespace on them with the components module and if that would 'hide' them from drupal: https://www.drupal.org/project/components

See the part about component-libraries in the .info file.

sherakama: Example: https://github.com/stanford-earth/matson/blob/8.x-1.x/matson.info.yml#L30

stanford_basic.libraries.yml
```diff
@@ -105,6 +105,17 @@ basic:
...
+
+fontawesome:
+  version: 4.7.0
+  remote: https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css
+  license:
+    name: FontAwesome
+    url: http://fontawesome.io/license/
+  css:
+    base:
+      //maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css: {}
```
pookmish: Since this is new stuff, should we use FA 5? I'd like to start at the latest that we can.

sherakama: Yes!

libraries/decanter/templates/elements/links/links.html.twig
pookmish: I'll try that out but i have my doubts

stanford_basic.libraries.yml
pookmish: https://fontawesome.com/get-started for quick ref

libraries/decanter/templates/elements/links/links.html.twig
sherakama: @josephgknox,

If we are running in to file name collisions we might have to pivot the naming convention in Decanter. Drupal looks for `something.html.twig` perhaps we just change everything in decanter to `something.twig` and remove the `html` part. 

That could keep things pretty clean.

sherakama: For reference: https://www.drupal.org/node/2354645

pookmish: Event the following had the same result:
```
component-libraries:
  pookmish_stuff:
    paths:
      - libraries
```

pookmish: Just in case you were curious.
screen shot 2018-05-03 at 9 27 13 am

sherakama: haha, I was.

pookmish: I'd be pro `something.twig` the `.html` is really just a drupal thing afterall isn't it?

josephgknox: Thanks, you two. I will work to fix the FA version -- good call on that.

@sherakama are there implications with the pattern library and style guide if we move the .html? I suspect not, but want to be sure it doesn't conflict with what KSS is looking for? I guess so long as we tell the SCSS files where to grab the markup then we're OK. I can proceed with making that fix in Decanter.

sherakama: Drat, this behaviour is contrary to the documentation. See `templates/` heading on: https://www.drupal.org/docs/8/theming-drupal-8/drupal-8-theme-folder-structure

pookmish: but decanter isnt drupal stuff. so coudn't we do that

sherakama: @josephgknox,

Other than a Canadian tonne of find and replace, there is nothing I can think of that would cause ill effects to the KSS styleguide.

sherakama:
@pookmish 

It might be useful to create a `decanter` namespace anyhow that points to the libraries/decanter folder so we can use that in our templates.

pookmish:
Yeah i could see that being useful. Have the ability to call a decanter template from a pattern/drupal template

josephgknox:
FontAwesomne 5 has been added and the latest Decanter has been pulled in (which removed .html from the Twig and JSON files).

libraries/decanter/hello-joe.txt
```diff
@@ -1 +0,0 @@
-What's up Shea!
```
sherakama: ha, what is this still doing here.

josephgknox: Lol!

PR #64, opened 2018-05-08
added search and menu regions. display help region too.

pookmish:
# READY FOR REVIEW

# Summary
- Adds 2 regions to better control layout. This eliminates the need for decanter classes on the blocks since they can be used in the page template. allowing us to put any block in those areas and have it sized appropriately.
- Display the help region
- page template cleanup from old regions that don't exist.

# Needed By (Date)
- whenever

# Urgency
- low

# Steps to Test

1. checkout this branch
2. install a clean drupal
3. enable the theme
4. review the home page for block placements.

# See Also


templates/page.html.twig
```diff
@@ -1,61 +1,38 @@
...
+    <div id=" menu-region">
+      {{ page.menu }}
     </div>
-  </header>
-  <!-- /#header -->
-{% endif %}
+  {% endif %}
+</header>
 
-{% if main_menu or secondary_menu %}
```
sherakama: By removing these you are removing support for the default menu variables and will be required to place a menu as a block in a region.

templates/page.html.twig
```diff
@@ -1,61 +1,38 @@
...
+  <div id="help-region">
+    {{ page.help }}
+  </div>
 {% endif %}
 
-<!-- ______________________ MAIN _______________________ -->
-
-{{ page.breadcrumb }}
-
-{% if page.highlighted | render %}
```
sherakama: What you got against this region? This is a default region provided by core
https://api.drupal.org/api/drupal/core%21modules%21system%21templates%21page.html.twig/8.2.x

templates/page.html.twig
```diff
@@ -1,61 +1,38 @@
...
-  <div id="highlighted">{{ page.highlighted }}</div>
-{% endif %}
-
-{{ title_prefix }}
-
-{% if title %}
-  <h1 class="title">{{ title }}</h1>
```
sherakama: Ooof. Are you sure you want to remove the page title from all pages and require it be placed explicitly for every page? This change gives me the willies.

templates/page.html.twig
```diff
@@ -1,61 +1,38 @@
...
-{% endif %}
-
-{{ title_suffix }}
-{{ page.help }}
-
-{% if tabs %}
-  <div class="tabs">{{ tabs }}</div>
```
sherakama: This would remove the edit/revision/delete/etc tabs from nodes and such. Are you sure?

templates/page.html.twig
```diff
@@ -1,61 +1,38 @@
...
-{% endif %}
-
-{% if action_links %}
-  <ul class="action-links">{{ action_links }}</ul>
```
sherakama: This is a secondary menu to the tabs and is less used but they provide actionable items and contrib modules might expect them to be there.

templates/page.html.twig
sherakama: I would do a quick check to see if any other modules expect it to be available and place stuff in it.

templates/page.html.twig
pookmish: These were set from within this template from respective regions within the page https://github.com/SU-SWS/stanford_basic/pull/64/files#diff-5d225d2e3e8538dcc0aa6138eae8454aL2
And dont we place the menu with a block anyways?

templates/page.html.twig
pookmish: But our theme doesnt define the region so it doesnt exist

templates/page.html.twig
pookmish: I thought we were placing the title with a block

templates/page.html.twig
pookmish: same as title

templates/page.html.twig
pookmish: same as title

pookmish:
Most of the changes in the template were just removing unused parts. If you review any of the core themes none of them have these parts like `{{ title }}`. All of theme only render out `{{ page.region_name }}`

templates/page.html.twig
sherakama: We do, and I am fine with removing this as long as that is the consensus on how this should work.

pookmish: these regions arent in the info.yml https://github.com/SU-SWS/stanford_basic/blob/8.x-2.x/stanford_basic.info.yml#L18

pookmish: I think this is same as the below discussion ref: 
> Does the theme have to define it if it is defined in the system module

stanford_basic.info.yml
```diff
@@ -11,6 +11,8 @@ libraries:
 regions:
   page_top: 'Page top'
   header: Header
+  search: Search
+  menu: Menu
   highlighted: Highlighted
```
sherakama: yank this out.

pookmish:
I removed highlighted region and changed skip nav text. the anchor was actually correct.

stanford_basic.info.yml
pookmish: yank this out? i thought we wanted this? we did it on humsci theme

PR #65, opened 2018-05-23
updated footer links for GDRP & translate the text

pookmish:
# READY FOR REVIEW

# Summary
- GDRP Links
- Maybe later we can make this more dynamic by use of a config.

# Needed By (Date)
- FRIDAY

# Urgency
- HIGH

# Steps to Test

1. checkout this branch
2. clear caches
3. validate new links are in the footer.

- MRC

# See Also

### DIFF
--- a/templates/components/global-footer.html.twig
+++ b/templates/components/global-footer.html.twig
@@ -15,16 +15,18 @@
     <div id="global-footer__info" class="decanter-width-five-sixths">
       <div id="global-footer__nav" role="navigation">
         <nav id="nav__global-links">
-          <a href="http://www.stanford.edu">Stanford Home</a>
-          <a href="http://visit.stanford.edu/plan/maps.html">Maps &amp; Directions</a>
-          <a href="http://stanford.edu/search/">Search Stanford</a>
-          <a href="http://emergency.stanford.edu/">Emergency Info</a>
+          <a href="https://www.stanford.edu">{{ 'Stanford Home'|t }}</a>
+          <a href="https://visit.stanford.edu/plan/">{{ 'Maps & Directions'|t }}</a>
+          <a href="https://www.stanford.edu/search/">{{ 'Search Stanford'|t }}</a>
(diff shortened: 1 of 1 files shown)


templates/components/global-footer.html.twig
```diff
@@ -15,16 +15,18 @@
     <div id="global-footer__info" class="decanter-width-five-sixths">
       <div id="global-footer__nav" role="navigation">
         <nav id="nav__global-links">
-          <a href="http://www.stanford.edu">Stanford Home</a>
-          <a href="http://visit.stanford.edu/plan/maps.html">Maps &amp; Directions</a>
-          <a href="http://stanford.edu/search/">Search Stanford</a>
-          <a href="http://emergency.stanford.edu/">Emergency Info</a>
+          <a href="https://www.stanford.edu">{{ 'Stanford Home'|t }}</a>
```
sherakama: You've added title attributes to the policy links do they not belong here as well?

pookmish: They arent on Scott's code snippet http://identity.stanford.edu/web-mobile.html#build-something-new

sherakama:
Themes don't really have the full fledged capabilities as modules in regards to installation. You get the config/install directory but you don't get a stanford_basic.install file FWIK. I think I would much rather have a drop and replace template file than a database update to change the links for the footer.

templates/components/global-footer.html.twig
sherakama: So they arn't 🤔 

I'll follow up with that. This is GTG.

sherakama: From John H.

> great question…we just made those updates yesterday and I know the title attribute is there on the second row of links because the visible text is an abbreviation of the full link, but let me doublecheck that the first row shouldnt have them too

sherakama: > just doublechecked…yes that was intentional because of the second row names being abbreviated. First row is already sufficiently descriptive in the displayed text

PR #71, opened 2018-08-09
set images without an alt to get a aria role instead

pookmish:
# READY FOR REVIEW

# Summary
- Images without an alt get the `role="presentation"` to indicate it is a decorative image.

# Needed By (Date)
- someday

# Urgency
- low

# Steps to Test

1. checkout this branch
2. clear cache
3. add an image via a field/media to a page without an alt
4. ensure that page has `role="presentation"` attribute.

# See Also

### DIFF
--- a/stanford_basic.theme
+++ b/stanford_basic.theme
@@ -134,3 +134,13 @@ function stanford_basic_preprocess_views_view(&$variables) {
 function _stanford_basic_change_characters($string) {
   return preg_replace("/[^a-zA-Z\d\s]/", '-', $string);
 }
+
+/**
+ * Implements hook_preprocess_image().
+ */
+function stanford_basic_preprocess_image(&$vars) {
+  // Decorative images get the role="presentation" attribute.
+  if (!isset($vars['attributes']['alt'])) {
+    $vars['attributes']['role'] = 'presentation';
+  }
+}


stanford_basic.theme
```diff
@@ -134,3 +134,13 @@ function stanford_basic_preprocess_views_view(&$variables) {
 function _stanford_basic_change_characters($string) {
   return preg_replace("/[^a-zA-Z\d\s]/", '-', $string);
 }
+
+/**
+ * Implements hook_preprocess_image().
+ */
+function stanford_basic_preprocess_image(&$vars) {
```
sherakama: 🤔 hmmm.. Going to have to think about this one a little bit. 

This is assuming that images intentionally don't have the alt value set. Is it true that alt is set but empty for those images that don't want an alt?

sherakama: 🤔 hmmm.. Going to have to think about this one a little bit. 

This is assuming that images intentionally don't have the alt value set. Is it true that alt is set but empty for those images that don't want an alt?

pookmish: at least with the media stuff, to have an empty alt part of the array, a user would have to intentionally populate the alt text field with a space. otherwise the alt text key will not exist.

pookmish: at least with the media stuff, to have an empty alt part of the array, a user would have to intentionally populate the alt text field with a space. otherwise the alt text key will not exist.

sherakama: Form validation, for the most part, requires alt text...

Alright, I think this fixes more issues than it creates. I'll bite.

pookmish: https://media.giphy.com/media/Z15pn0hiqA7uw/giphy.gif

PR #92, opened 2019-10-18
D8CORE-631: added max-width for h2 and h3

cjwest:
# READY FOR REVIEW

# Summary
_See ticket:  https://stanfordits.atlassian.net/browse/D8CORE-631_
This adjusts the max-width for H2 and H3 to be proportional to the body text.

# Needed By (Date)
- Sprint end

# Urgency
- Fixes broken stuff

# Steps to Test

Checkout this branch
Create a page with:
- H2 and body text
- H2 centered and body text
- H3 and body text
Verify you like how it looks.

- D8Core (https://github.com/SU-SWS/acsf-cardinalsites)

## Related JIRA ticket(s)
https://stanfordits.atlassian.net/browse/D8CORE-631

## More Information
This problem occurs because of how we are limiting the number of characters per line.

## Folks to notify
@kerri-augenstein

# See Also


src/scss/components/typography/_line_length.scss
```diff
@@ -5,22 +5,17 @@
 // Set a maximum line length.
 // This sets a maximum width based the size of the letter "O".
 
-h4 {
-  max-width: 68ch;
-}
-
-h5 {
+li, p {
```
cjwest: @sherakama,
I added the <li>. Let me know if that's okay. I think it is covered by the what this work intends to accommplish.

cjwest:
@kerri-augenstein,

Here's how this looks:
staggered-max-line-length

I included limiting the line length for the <li> because I think it is used fairly frequently. Let me know if there's anything else.

src/scss/components/typography/_line_length.scss
sherakama: Unfortunately, this did not work as it messes with the main navigation and other navigation-related items.

src/scss/components/typography/_line_length.scss
```diff
@@ -5,22 +5,17 @@
...
   max-width: 70ch;
 }
 
-h6 {
-  max-width: 80ch;
+p.text-align-right {
```
sherakama: Let's pull this out of the generic typography component and create a WYSIWYG component so you can nest these selectors under `.su-wysiwyg-text` in order to encapsulate them.

src/scss/components/typography/_line_length.scss
sherakama: We might be able to keep the `<li>` if you move them to a WYSIWYG component and nest the selectors under `.su-wysiwyg-text`

cjwest:
@sherakama,
I agree with your suggestion. We can limit line lengths as we find the and reduce possible global weirdness.
Let me know if this works.

PR #101, opened 2019-11-16
D8CORE-851 Fix margin for figures

pookmish:
# READY FOR REVIEW

# Summary
- Override browser styles for `<figure>` tags

# Need Review By (Date)
- eventually

# Urgency
- low

# Steps to Test
1. checkout this branch
1. add a large image in the WYSIWYG, add a caption to it and align it to the left
1. add a blob of text below or above it
1. save and verify the image left alignment is the same as the text left alignment

# See Also


src/scss/base/_typography.scss
```diff
@@ -39,3 +39,7 @@
 .su-quote-text {
   @include type-b;
 }
+
+figure {
+  margin: 0;
```
sherakama: This is already in Decanter. (https://github.com/SU-SWS/decanter/blob/master/core/src/scss/elements/image/_figure.scss)

Why is it needed in stanford_basic?

pookmish: ahh why indeed.... :thinking:

sherakama: close?

pookmish: we could if we can figure out why its not added to the site right now

sherakama: https://github.com/SU-SWS/decanter/tree/6.0.0-alpha.1/core/src/scss/elements/image

The figure partial is missing in 6.0.0-alpha.1

Either a new release needs to be punched or just wait.

pookmish: works for me. it'll fix itself later

PR #111, opened 2020-01-22
D8 core 941 margin top

cjwest:
# READY FOR REVIEW 

# Summary
_See ticket:  https://stanfordits.atlassian.net/browse/D8CORE-941_

# Needed Review By (Date)
- Sprint end

# Urgency
Needs to get done
 
# Steps to Test

- Checkout this branch
- npm run publish ; lando drush cr
- Verify Check all wysiwyg headings and styles with figma: https://www.figma.com/file/mQ5izx1G6f8DEUkgwWIHWP/D8-End-User-Display?node-id=0%3A1
Check Drop cap on both Chrome and Firefox

- D8Core (https://github.com/SU-SWS/acsf-cardinalsites)

# QA

## Cross-Browser testing:
What browsers have been tested?
- Chrome - Yes
- Firefox - Yes
- Safari
- Internet Explorer
- Android (Edge)
- iPhone (Safari)

## Unit Testing

## Related JIRA ticket(s)
https://stanfordits.atlassian.net/browse/D8CORE-941

## Folks to notify
`@sherakama` 
`@kerri-augenstein`

# See Also


sherakama:
@cjwest 2 code climate fixes

src/scss/base/_typography.scss
```diff
@@ -5,103 +5,122 @@
  * See: node_modules/decanter/core/src/scss/elements/typography/_typography.scss
  */
 
-// Caption Styles.
-.caption {
-  figcaption {
-    text-align: center;
+.su-wysiwyg-text {
+
+
```
sherakama: extra line not needed.

src/scss/base/_typography.scss
```diff
@@ -5,103 +5,122 @@
...
+  //
+  // Headings overrides.
+  //
+  h1 {
+    @include modular-spacing('margin', 4 null 5 null);
```
sherakama: I'd rather not have top margins, especially on headers. This is going to make it exceptionally difficult to top align a heading in any container and I am not interested in adding overrides for every specific container.

src/scss/base/_typography.scss
```diff
@@ -5,103 +5,122 @@
...
-}
+  // Caption Styles.
+  .caption {
+    figcaption {
+      text-align: center;
+    }
+  }
+  p {
+    @include modular-spacing('padding-bottom', 1);
+    margin-bottom: 0;
```
sherakama: This should use the mixin `@include margin(null null 0)`

src/scss/base/_typography.scss
```diff
@@ -5,103 +5,122 @@
...
     font-weight: bold;
-    font-size: 3em;
-    line-height: 0.7;
-    margin: 18px 12px 0 0;
-    text-transform: uppercase;
+    line-height: 33px;
+    width: 80%;
+
+    &:first-of-type {
+      padding-top: 0;
```
sherakama: use padding mixin.

src/scss/base/_typography.scss
```diff
@@ -5,103 +5,122 @@
...
+      margin-top: 0;
```
sherakama: Use margin mixin

src/scss/base/_typography.scss
```diff
@@ -5,103 +5,122 @@
...
-}
+  // Drop cap styles.
+  .su-drop-cap {
+    &::first-letter {
+      float: left;
+      font-style: normal;
+      font-weight: bold;
+      font-size: 3.5em;
+      line-height: .6;
+      margin: 11px 12px 0 0;
```
sherakama: use margin mixin

src/scss/base/_typography.scss
```diff
@@ -5,103 +5,122 @@
...
+
+    // Adding a drop-cap will automatically apply intro-text
+    @include intro-text;
+  }
 
-.su-intro-text {
-  @include modular-spacing('margin-bottom', 3);
-}
+  .su-font-splash {
+    @include modular-spacing('margin', -1 null 2 null);
```
sherakama: Negative top margin? Why?

src/scss/base/_typography.scss
```diff
@@ -5,103 +5,122 @@
...
+  .su-quote-text {
+    @include type-b;
+  }
+
+  // A shadowy type.
+  .su-related-text {
+    @include box-shadow('paper');
+    @include modular-spacing('margin-bottom', 3);
+    border: 1px solid #e3e3e3;
+    padding: 25px;
```
sherakama: use padding mixin.

sherakama:
This needs an in-person discussion. @kerri-augenstein & @cjwest please book some time with me to discuss the approach on margins.

src/scss/base/_typography.scss
cjwest: @sherakama, What would you suggest for creating space between a <p> and a <h2>?

@kerri-augenstein FYI

src/scss/base/_typography.scss
cjwest: Thanks!

src/scss/base/_typography.scss
cjwest: @kerri-augenstein has it specified here:
https://www.figma.com/file/mQ5izx1G6f8DEUkgwWIHWP/D8-End-User-Display?node-id=0%3A1

src/scss/base/_typography.scss
cjwest: Thanks!

src/scss/base/_typography.scss
cjwest: Got it! Thanks.

src/scss/base/_typography.scss
cjwest: Thanks!

src/scss/base/_typography.scss
cjwest: Thanks!

src/scss/base/_typography.scss
sherakama: Oh right, -1 mean less spacing not negative margins eg: (-10px). Got it.

sherakama:
@cjwest 

From our slack conversation with the group. This work will need to be revised to follow the following conventions.

1. The outer spacing should only be bottom directional
2. The inner spacing can be multidirectional when the component has known elements
3. The spacing can be multidirectional but should be bottom only where possible when the elements inside a component can be arranged in an unknown order.
4. For the case where the component elements do not have a known arrangement force the first element under the component wrapper to have no top margin (when there is no background) and the last element to have no bottom margin (always). This will allow the spacing of the components to stay rigid and the elements with to change.

For this ticket, you will need to enforce the top and bottom element margins and then create sibling selectors for all the possible combinations of headings/paragraphs to ensure the "chunks" are correct.

cjwest:
@sherakama,
I refactored the headings and removed spacing after the last <p>.
Let me know if there's anything more for this PR!

cjwest:
Here's the ticket for re-thinking and re-factoring margins:
https://stanfordits.atlassian.net/browse/D8CORE-1203

src/scss/base/_typography.scss
```diff
@@ -5,103 +5,150 @@
...
+  // Headings overrides.
+  //
+
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6 {
+    // Remove margin-top and padding-top when a heading is the first item on the page
```
sherakama: It is not just headings that need to have the first element have the margin removed. It is only the first element of any type. We don't want to remove the top margin of just the headings but also p, img, figure, or whatever is added as the first item.

I think the selector and code you are looking for in this case is 
```
.su-wysiwyg-text {
  &:first-child {
    @include margin(0 null null null);
    @include padding(0 null null null);
  }
}
```

src/scss/base/_typography.scss
```diff
@@ -5,103 +5,150 @@
...
+  }
 
+  // A shadowy type.
+  .su-related-text {
+    @include box-shadow('paper');
+    @include modular-spacing('margin-bottom', 3);
+    @include padding(25px);
+    border: 1px solid #e3e3e3;
+
+    // Yeah, let's not remove this.
```
sherakama: What is this comment all about?

src/scss/base/_typography.scss
```diff
@@ -5,103 +5,150 @@
...
+    @include margin(0 null null null);
+
+    // Remove margin-top and padding-top when a <p> is the first item on the page
+    &:first-of-type {
+      @include margin(0 null null null);
+      @include padding(0 null null null);
+    }
+
+    // Remove margin-bottom and padding-bottom when a <p> is the last item on the page
+    &:last-of-type {
```
sherakama: Elements other than `<p>` can be the last thing on the page. Refactor this so that you are removing margin and padding to the last element no matter what it is.

sherakama:
@cjwest ping

src/scss/base/_typography.scss
cjwest: If you  remove the margin-bottom from the related-text, it doesn't look right

src/scss/base/_typography.scss
cjwest: @sherakama,
Hmmm, I don't think that this is how `first-child` works.

We can have:
```
.su-wysiwyg-text {

    p:first-child {
      @include margin(0 null null null);
      @include padding(0 null null null);
    }
```

Which is very similar to 
```
p {
    // Remove margin-top and padding-top when a <p> is the first item on the page
    &:first-of-type {
      @include margin(0 null null null);
      @include padding(0 null null null);
    }
}
```

src/scss/base/_typography.scss
sherakama: Can you fix the comment to reflect your explanation, please.

src/scss/base/_typography.scss
sherakama: Go with the approach that works for you. What I am looking for is that your selector picks up the first child element of su-wysiwyg-text no matter what html element it is. If you can use a psuedo selector or * selector to do the job please do so.

cjwest:
@sherakama,
I refactored the first and last child selectors, and added further specificity to some of the selectors.
Let me know what you think!

src/scss/base/_typography.scss
```diff
@@ -5,103 +5,157 @@
  * See: node_modules/decanter/core/src/scss/elements/typography/_typography.scss
  */
 
-// Caption Styles.
-.caption {
-  figcaption {
-    text-align: center;
+.su-wysiwyg-text {
```
sherakama: In retrospect, these styles should be coupled to paragraph type and not this theme. These look good but it is time to move them to https://github.com/SU-SWS/stanford_paragraph_types

sherakama:
The styles look great. Thanks @cjwest. As this code as matured we have scoped it to a paragraph type and now need to move it out of the theme and in to the profile so that the styles follow the entity type. I've opened a new PR there for you to review. Once that is merged we can remove the styles from this theme and merge this branch.

https://github.com/SU-SWS/stanford_profile/pull/75

Thank you for this work and all of the revisions.

sherakama:
https://github.com/SU-SWS/stanford_profile/pull/75 was merged. This is GTG.

sherakama:
ckeditor changed. Wut?

cjwest:
Should this PR be closed @sherakama ?

sherakama:
I'd like to check and see what we are changing with the ckeditor styles first. We may need to patch those in again.

sherakama:
Drat it looks like the ckeditor was using those styles in typography. I'll have to move those in to the profile as well. This PR is good to merge and I'll pick back up on the editor styles in stanford_profile.

PR #114, opened 2020-01-23
D8CORE-1012 Center images in the wysiwyg

pookmish:
# READY FOR REVIEW

# Summary
- Centered images in the wysiwyg needed some css to make them centered.

# Need Review By (Date)
- 1/29

# Urgency
- low

# Steps to Test
1. checkout this branch
1. create a page with a wysiwyg
1. add a small image and center align it
1. after saving, validate the image is centered.

# See Also


sherakama:
Two CC fixes and the CSS selectors should be encapsulated within `.su-wysiwyg-text` so that they are not effecting everything globally.

pookmish:
:+1:

src/scss/components/wysiwyg/_media.scss
```diff
@@ -0,0 +1,10 @@
+@charset 'UTF-8';
+
+.su-wysiwyg-text {
```
sherakama: The more that I think of this the more I think this should be moved to the profile as the theme shouldn't know about our components.

sherakama: I'll make a ticket for later.

PR #131, opened 2020-02-14
D8 core 1286

jenbreese:
# READY FOR REVIEW

# Summary
- Fixing the <h1> spacing on the search page.

# Review By (Date)
- 2/14

# Urgency
- med

# Steps to Test

1. Pull in the branch.
2. Clear caches.
3. Search
4. Verify the <h1> on the search results pages has a margin top that is the same as our other pages.

- na

- D8CORE-1286
- @kerri-augenstein 

# See Also


src/scss/components/site-search/_site-search.scss
```diff
@@ -0,0 +1,77 @@
+@charset "UTF-8";
+
+//
+// Custom Styles For Stanford Page Search form and search results.
+//
+// Search Header
+h1 {
```
sherakama: H1's cannot live out on their own or this will adjust all h1s everywhere. Is this just for the search page? If so, you will either need to nest the selector for this item or you can create a separate stylesheet that is only added on the search page.

jenbreese: I thought because it was on the _site-search.scss it was only on the search page. I'll fix.

jenbreese:
I corrected the one you mentioned. I looked at the file again and saw the same problem with the ol so I fixed that one too.

src/scss/components/site-search/_site-search.scss
```diff
@@ -4,6 +4,34 @@
 // Custom Styles For Stanford Page Search form and search results.
 //
 // Search form on search results page
+
+// Search Header
+.page-search-content {
```
sherakama: The practice of using a body selector is discouraged as it creates an extremely strong selector that will be hard to override later. I'm going to merge this for the release today but let's have a refactor session together and I will show you how to set up a template specific library using twig.

https://www.drupal.org/node/2456753

jenbreese: Great! I will check it out. Thank you.

PR #141, opened 2020-02-26
D8CORE-1327: fixing the header lockup to have special characters

jenbreese:
# READY FOR REVIEW

# Summary
- Currently, html entities are encoded and passed into the twig templates rendering & as &amp; This is particularly a problem for the header and footer's lockup.

# Review By (Date)
- Thursday 

# Urgency
- Med/Low

# Steps to Test

1. Set the site name to Bugs & 'Bunny'
2. Clear cache and review the local footer lockup
3. See double encoding of the &
4. See the ' encoded.
5. Check out this branch and import config
6. Clear cache and see only the & and the '
6. Shea fixed up the double encoding in the config.

- D8CORE-1327

- D8CORE-1327
- https://github.com/SU-SWS/stanford_profile/pull/123
- https://github.com/SU-SWS/jumpstart_ui/compare/D8CORE-1327
- @sherakama 

# See Also

### DIFF
--- a/templates/block/block--system-branding-block.html.twig
+++ b/templates/block/block--system-branding-block.html.twig
@@ -20,16 +20,16 @@
 
 {# Do not default the line 1 on variant O, S, or T #}
 {% if lockup.option != "o" and lockup.option != "s" and lockup.option != "t" %}
-  {% set line1 = content.site_name|render|striptags %}
+  {% set line1 = content.site_name|render|raw %}
 {% endif %}
 
 {% if lockup.line1 is not empty %}
- {% set line1 = lockup.line1|render|striptags %}
+ {% set line1 = lockup.line1|render|raw %}
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


sherakama:
> I dont believe there's a need to do `|render|raw` if you aren't stripping tags. You could just pass the variable with that.

Let me check that out. It is looking like variables passed into twig templates with as a plain text render array are being autoescaped (which they should be). So the two options are to alter the render array before it hits twig or to mark it as 'safe' with `|raw` but I need to confirm that suspicion.

pookmish:
I confirmed if you remove `|render|raw` the result still appears to fix the issue.

jenbreese:
@pookmish I confirm you are right. :) That did work.

sherakama:
As these are not fields going into this template this works well.

We need the `striptags` filter on variables that are outputting fields as Drupal debugging adds all the HTML comments even though it is rendering plain text.

PR #146, opened 2020-03-13
D8CORE-1349 Added visual indicator for links to unpublished pages

pookmish:
This needs some front end attention.


pookmish:
I'm going to pass this off to @jenbreese 
Please let me know if you need anything from me on this.

jenbreese:
# READY FOR REVIEW

# Summary
- This styles the indicators pointing out the unpublished files.

# Review By (Date)
- Soonish

# Urgency
- Low

# Steps to Test

1. Pull in this branch.
2. npm run publish
3. Clear caches.
4. Log in.
5. Unpublish a page.
6. Verify you get the the indicator.
7. Try on on main and secondary navigation pages.

- D8CORE-1349

- D8CORE-1349

- @pookmish 

# See Also

pookmish:
Cool. thanks @jenbreese this looks good to me. @sherakama care to take a look?

src/scss/admin/_menu.scss
```diff
@@ -0,0 +1,37 @@
+@charset "UTF-8";
+
+nav {
+  .unpublished-indicator {
+    @include padding(4px 10px);
+
+    content: "Unpublished Page";
```
sherakama: It is not recommended to add text in this manner as it doesn't get translated. In this case, it doesn't matter due to the authenticated user requirement, but for future reference, added text through CSS has a negative impact on SEO and accessibility.

src/scss/admin/_menu.scss
```diff
@@ -0,0 +1,37 @@
+@charset "UTF-8";
+
+nav {
```
sherakama: I don't think you need to next the unpublished-indicator in a nav selector. You can get away with just the css class selector.

templates/menus/macros/nav-menu.twig
```diff
@@ -28,7 +28,12 @@
     {% endif %}
 
     <li{{ list_attributes }}>
-      <a{{ link_attributes }}>{{ item.title }}</a>
+      <a{{ link_attributes }}>
+        {{ item.title }}
+        {% if item.unpublished %}
+          <div class="unpublished-indicator">Unpublished Page</div>
```
sherakama: Strings should be run through the `trans` filter.

templates/menus/macros/secondary-nav-menu.twig
```diff
@@ -30,7 +30,12 @@
     {# Skip non-trail items below the first level. #}
     {% if item.in_active_trail == true or menu_level == 1 or parent.in_active_trail == true %}
       <li{{ list_attributes }}>
-        <a{{ link_attributes }}>{{ item.title }}</a>
+        <a{{ link_attributes }}>
+          {{ item.title }}
+          {% if item.unpublished %}
+            <div class="unpublished-indicator">Unpublished Page</div>
```
sherakama:
```suggestion
            <div class="unpublished-indicator">{{ "Unpublished Page"|t }}</div>
```

jenbreese: Ok. I can do that.

src/scss/admin/_menu.scss
```diff
@@ -0,0 +1,37 @@
...
+      border-left: 10px solid transparent;
+      border-right: 10px solid transparent;
+      border-top: 10px solid $su-color-blue;
+      position: absolute;
+      bottom: -10px;
+      right: 90px;
+    }
+  }
+
+  &.su-secondary-nav {
```
sherakama: No need to nest this in a nav component and have a strong `nav.su-secondary-nav .unpublished-indicator` selector output. `.su-secondary-nav .unpublished-indicator` should suffice.

src/scss/admin/_menu.scss
jenbreese: Ok. I'll work on that tomorrow.

stanford_basic.theme
```diff
@@ -48,7 +48,7 @@ function stanford_basic_preprocess_html(&$variables) {
     if (!empty($alias)) {
       $name = str_replace('/', '-', $alias);
       $variables['attributes']['class'][] = 'page-' . $name;
-      list($section,) = explode('/', $alias, 2);
+      [$section,] = explode('/', $alias, 2);
```
sherakama: ooof. I can see where this short syntax can get a little ugly. That hanging comma bugs me.

src/scss/admin/_menu.scss
jenbreese: Ok. I'll take a look.

sherakama:
I really like the idea of this work as having a visual indicator to show which items are unpublished is super helpful. I left a few code comments above but also have a couple comments about the visual display. The placement of the tooltip seems obtrusive and jarring. The overlay and positioning appear way out of the natural flow of content and really catch my eye. Perhaps @kerri-augenstein has some feedback on the visual appearance of the notice. 

Screen Shot 2020-03-17 at 10 08 14 PM

Screen Shot 2020-03-17 at 10 08 30 PM

kerri-augenstein:
@sherakama, thanks for commenting on this!! I think what we're going for is for these pages to be really, really obvious in the display, like an alert. So I'm not too bothered that the tooltip is obtrusive, and as intended catches your eye, yey, goal! But. 

Maybe we could have this tooltip only appear for a duration of seconds, 3, or 5??? And then reappear on-hover? The link will be paired with a faint blue background behind the themselves as well, so that'll help extend the visual indication, after the tooltip has faded (though the light blue is light, we may need to test how visible this ends up being for our users).

jenbreese:
Ok, I fixed up most of the comments from @sherakama and @kerri-augenstein. I used an animation for the indicator to show up. I'm working on the hover but it isn't right yet. So thoughts on that appreciated.

jenbreese:
@sherakama or @pookmish want to take another look?

pookmish:
I like that animation and the fade away after some time. it helps to identify but isn't obtrusive to the experience.

sherakama:
This looks good. Just one visual oddity.

pookmish:
blame @sherakama https://github.com/SU-SWS/stanford_basic/blob/e911bc914300752f30488a5798f918e9ebb50cd7/templates/node.html.twig#L37

jenbreese:
@sherakama I thought that was intentional. Would you or @pookmish like me to style it?

pookmish:
@jenbreese i think it was intentional when Shea initially created the theme back in 2016 :laughing: 
```
$ git log -S node--unpublished
commit 2f3310139a077a60b00281efcfeef95c2fa6d8d7
Author: Shea McKinney <sherakama@gmail.com>
Date:   Wed Dec 28 15:29:11 2016 -0800

    Initial 8.x Commit
```
I think we might as well style it as part of this.

jenbreese:
@pookmish Ok. I'll do that.

jenbreese:
@kerri-augenstein  I'd like to style this 'unpublished page" text. How would you like it to look?
image

sherakama:
haha yeah, that is a relic. 

The text doesn't have to stay. Not sure what my intentions with that were. Perhaps it could be swapped out with a yellow alert banner "This page is currently unpublished and not visible to the public."

pookmish:
i like the alert banner idea. use what they are already some what familiar with to display the information.

jenbreese:
@pookmish @sherakama I updated the text and added banner styling.

templates/node.html.twig
```diff
@@ -34,7 +34,7 @@
       {% endif %}
 
       {% if not node.published %}
-        <p class="node--unpublished">{{ 'Unpublished'|t }}</p>
+        <p class="node--unpublished su-alert--warning">{{ 'This page is currently unpublished and not visible to the public.'|t }}</p>
```
sherakama: The alert template is available, why not use it?

```
{% 
set data = {
    'attributes': create_attribute({'class': 'su-alert su-alert--warning'}),
    'alert_text': "My message"|t,
  }
%}
{% include "@basic-dist/decanter/components/alert/alert.twig" with data %}
```

jenbreese: Indeed sir! :) all done.

src/scss/admin/_menu.scss
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// Will need to make this less specific.
+nav {
+  li:hover {
+    .unpublished-indicator {
+      opacity: 1;
+    }
+  }
+}
```
sherakama: I think this is fine. Not much you can do about that.

pookmish: should it be `a:hover`. with this the indicator will show on all children of that link and all indicators of those children

sherakama: That is a good point. I think you are probably right. 

Also, this doesn't appear to be working at all on hover at the moment.

jenbreese: Yes, I was just about to write that.  I'll fix.

jenbreese:
@sherakama and @pookmish I've fixed the hover.

src/scss/admin/_menu.scss
```diff
@@ -0,0 +1,53 @@
...
+  100% { opacity: 0; }
+}
+
+a:hover {
+  .unpublished-indicator {
+    display: inline-block;
+    opacity: 1;
+    animation: fade 5s ease-in-out;
+  }
+}
```
sherakama:
```suggestion
a:hover,
a:focus {
  & > .unpublished-indicator {
    display: inline-block;
    opacity: 1;
    animation: fade 5s ease-in-out;
  }
}
```

sherakama: Perhaps include the focus state and ensure it is only the direct unpublished indicator?

pookmish: IMO the direct decedent isn't necessary and increases the amount of specificity that might not be wanted if we ever want to override the styles with a subtheme.

sherakama: That is a fair point. With the activation on the `a` tag now it probably isn't needed and the less specificity is valued.

jenbreese: Got it.

sherakama:
Thank you.

PR #148, opened 2020-03-26
D8CORE-1073: A new way to move the search box into the mobile menu

jenbreese:
# READY FOR REVIEW

# Summary
- This is so the sites can have a search box in the mobile menu. 

# Review By (Date)
- We want it to be in v. 1.1.0

# Urgency
- important

# Steps to Test

1. Pull in branch.
2. In your local site, add an additional search block to the Menu under the Main Navigation
3. Build and refresh search too.

# Expectations
1. You should see only one search box when your window is open large. 
2. On a mobile view you should only have one search box at the top of the open menu.
3. Verify that the submenus do not have a search box.
4. Verify the search boxes work.
5. Note: I wrapped the search <div> with an <li> and added it to the menu. This means it inherits all the styles. It is valid markup to have a <div> in an <li>. 
6. Note 2: I added a wildcard selector to select and hide other search boxes added in the block layout. I'm not sure its necessary. 
7. Note 3: I need some assistance with the Accessibility testing. 

- I think it would be unnoticed change on live sites since they'd have to add the search block. 

- D8CORE-1073
- @kerri-augenstein , I think you'd want to review as well.

# See Also


src/js/stanford_basic.behavior.js
```diff
@@ -25,6 +25,11 @@ export default {
       if (!$sn) {
         $('.su-skipnav--secondary', context).remove();
       }
+
+      // Move the second search block to the navigation. Hide it and then only show for mobile sites.
+      var $ms = $('.su-site-search[data-drupal-selector="search-block-form-2"]');
+      $($ms).prependTo('nav.su-multi-menu button + ul').wrap('<li></li>');
```
sherakama: Cheeky but I like it. 

The block form id can change so your selector for `$ms` is going to have to change. We cannot rely on `search-block-form-2`. You can assume we want the search form in the masthead for this and write a descendant selector for the search box. 

I don't think we can assume there will only ever be one `su-multi-menu` on a page. Perhaps change the `'nav.su-multi-menu button + ul'` selector to `.su-masthead .su-multi-menu > ul`

src/scss/components/multi-menu/_multi-menu.scss
```diff
@@ -257,3 +257,20 @@
     }
   }
 }
+
+// Search box styling for devices and mobile views.
+// -----------------------------------------------------------------------
+div[data-drupal-selector="search-block-form-2"],
```
sherakama: we cannot rely on this id. Please refactor.

jenbreese:
@sherakama I fixed the specificity. Can you point me the right way on how to manage the keyboard navigation? In the mobile view, tabbing skips the mobile search. But on the phone, you aren't tabbing.

src/js/stanford_basic.behavior.js
```diff
@@ -25,6 +25,11 @@ export default {
       if (!$sn) {
         $('.su-skipnav--secondary', context).remove();
       }
+
+      // Move the second search block to the navigation. Hide it and then only show for mobile sites.
+      var $ms = $('nav + .su-site-search');
+      $($ms).prependTo('.su-masthead .su-multi-menu > ul').wrap('<li></li>');
```
sherakama, line 31:
```suggestion
      var $ms = $('.su-masthead .su-site-search', context);
      if ($ms.length == 1) {
        $($ms).prependTo('.su-masthead .su-multi-menu > ul', context).wrap('<li></li>');
      }
```

A couple of changes here. When in the `attach` behavior in Drupal you should be using the `context` option within you jquery query statements. Otherwise, this function will trigger repeatedly each time an ajax call happens. The context bounds the code to run only on the content that is changing. We also cannot assume that all sites going forward will have the search box in their masthead. We should check for it.

src/scss/components/multi-menu/_multi-menu.scss
```diff
@@ -257,3 +257,33 @@
...
+    .su-site-search {
+      display: none;
+    }
+  }
+}
+
+.su-multi-menu {
+  // show the extra search box in responsive menu.
+  @include grid-media-between('xs', 'md') {
+    div[data-drupal-selector='search-block-form-2'] {
```
sherakama, line 283:
```suggestion
    .su-site-search {
```

search-block-form-2 doesn't exist for me.

sherakama:
> @sherakama I fixed the specificity. Can you point me the right way on how to manage the keyboard navigation? In the mobile view, tabbing skips the mobile search. But on the phone, you aren't tabbing.

The tab index seems ok overall. I _think_ when on mobile after the user clicks/taps on the hamburger menu to expand the menu that the focus should be in the search bar but perhaps not as that is not what we have on the https://decanter.stanford.edu/component/composite-main-navigation/ page.

src/scss/components/multi-menu/_multi-menu.scss
jenbreese: Oh man. I'm so sorry. :( I'll fix.

sherakama:
> Maybe I should check with @cjwest?

This might be a Jiatyan question. I think the focus on the first link item is not too bad overall and we can catch up later if both of those two are busy.

src/js/stanford_basic.behavior.js
```diff
@@ -25,6 +25,15 @@ export default {
       if (!$sn) {
         $('.su-skipnav--secondary', context).remove();
       }
+
+      // Check for search box and move the second block to the mobile navigation.
+      // Hide it and then only show for mobile sites.
+      var $search = $('#block-stanford-basic-search', context).length;
```
sherakama, line 31: There isn't really a guarantee that this block is available, or even, the right one. It is possible that someone could put this block in the sidebar &/or the header &/or the footer. I think I'd rather this selector be looking at a specific region or within the masthead for the `.su-site-search` class than the ID. Drupal is funny this way.

src/js/stanford_basic.behavior.js
```diff
@@ -25,6 +25,15 @@ export default {
...
+      if ($search) {
+        var $ms = $('nav + .su-site-search', context);
```
sherakama, line 33: This selector needs to be more specific to target the main navigation. It is possible that someone could put a su-site-search in the sidebar nav and you wouldn't want to adjust that.

src/js/stanford_basic.behavior.js
jenbreese: Ok. I thought the block id was predicable no matter it was put. I don't mind changing; I just thought it was a safe choice.

jenbreese:
I changed the specificity and also tried out the search block in the footer and article body. It worked as expected. I wasn't able to simply add the search box to the side bar without major modifications. I think that was by design, right?

src/js/stanford_basic.behavior.js
```diff
@@ -25,6 +25,15 @@ export default {
...
         $('.su-skipnav--secondary', context).remove();
       }
+
+      // Check for search box and move the second block to the mobile navigation.
+      // Hide it and then only show for mobile sites.
+      var $search = $('.su-masthead .su-site-search', context).length;
+      if ($search) {
+        var $ms = $('.su-masthead nav + .su-site-search', context);
+        $($ms).prependTo('.su-masthead .su-multi-menu > ul', context).wrap('<li></li>');
+      }
```
sherakama, line 35:
```suggestion
      var $search = $('.su-masthead .su-site-search', context);
      if ($search.length) {
        var $searchcopy = $search.clone();
        $searchcopy.prependTo('.su-masthead .su-multi-menu > ul', context).wrap('<li></li>');
      }
```

Instead of adding a second search block to the same region how about just copying the first one that is there?

sherakama: Oh, IDs become an issue.

sherakama: Can't do that unless we fix the IDs as well. Drat.

sherakama:
I need to circle back with you on this. Sorry for the delay.

src/js/stanford_basic.behavior.js
```diff
@@ -26,6 +26,17 @@ export default {
...
+      // Check for search box and move the second block to the mobile navigation.
+      // Hide it and then only show for mobile sites.
+      var $search = $('.su-masthead .su-site-search', context).length;
+      if ($search) {
+        var $ms = $('.su-masthead .su-site-search', context);
+        $($ms.clone())
+          .prependTo('.su-masthead .su-multi-menu > ul', context)
+          .wrap('<li class="su-mobile-site-search"></li>')
+          .attr('id', 'block-stanford-basic-search-mobile');
+      }
```
pookmish, line 38:
```suggestion
          .attr('id', 'block-stanford-basic-search-mobile')
          .find('[id]').each(() => {$(this).attr('id', $(this).attr('id') + '-mobile')});
      }
```

jenbreese: Ok. I updated it. I've tested on Chrome and Firefox. Going to double check on IE/browserstack.

sherakama (approved):
Looks and works great. Thanks for getting to this.

PR #150, opened 2020-04-08
D8 core 1499

jenbreese:
# READY FOR REVIEW

# Summary
- adjustments for the mulitparagraphs per row.

# Review By (Date)
- soonish

# Urgency
- high

# Steps to Test

1. Set up 4 per row: https://github.com/SU-SWS/stanford_profile/pull/159 : checkout that branch and run blt drupal:update and that'll give you the ability to have 4 per row
2. Pull in this branch.
3. Start adding content.
4. Try it out on the homepage/full width page
5. Try it on an interior page.
6. Try it on a page with a secondary navigation.

# Notes
1. I've tried it on Chrome/Mac, FF/Mac, Safari/Mac, IE, Edge.  There are some issues with FF on Mac and PC (browserstack)
2. I tried to not change any margins. I shrunk some set sizing when it overran out of the browser. In a design meeting, we thought stacking 4 across starting at LG would be helpful since it the paragraphs are so small.  

- https://github.com/SU-SWS/stanford_profile/pull/159

- D8CORE-1499
- https://github.com/SU-SWS/stanford_profile/pull/159
- @sherakama 

# See Also


stanford_basic.theme
```diff
@@ -317,3 +317,17 @@ function stanford_basic_pattern_process_contextual(array $element) {
...
+/**
+ * Implements hook_preprocess_HOOK().
+ * for adding grid classes to multirow paragraphs
+ */
+function stanford_profile_styles_preprocess_react_paragraphs(&$variables) {
+  foreach ($variables['rows'] as &$row) {
+    foreach ($row['items'] as &$item) {
+      $width = $item['width'];
+      $item['attributes']->addClass("flex-$width-of-12");
+      $row['attributes']->addClass("container-$width-columns");
```
pookmish, line 330: you are adding multiple classes on the row. im' curious why
so the result could end up being 
`<div class="container-3-columns container-6-columns container-3-columns">`

src/scss/components/wysiwyg/_multi_per_row.scss
```diff
@@ -0,0 +1,62 @@
...
+      }
+
+      // Setting the max width to help with display and
+      // to keep the large margins on full-width and the
+      // homepage
+      @include grid-media('xl') {
+        .ptype-stanford-card.flex-3-of-12,
+        .ptype-stanford-media-caption.flex-3-of-12,
+        .ptype-stanford-wysiwyg.flex-3-of-12 {
+          max-width: 25%;
```
pookmish, line 19: shouldn't these already get max widths  or width limits somehow from the `flex` class?

jenbreese: They do have max-widths and they are too big. So that row pushes out the content from the container.

stanford_basic.theme
jenbreese: Oh I hadn't seen it add multiple classes per row. I wanted the class on the row so I change to the display: block from flex at the lg breakpoint. What steps do I do to see the multiples?

jenbreese:
@pookmish, can you try this version out? I moved the container class. Does that solve your concern? I also simplified and condensed what I was doing in the Sass.

stanford_basic.theme
```diff
@@ -317,3 +317,17 @@ function stanford_basic_pattern_process_contextual(array $element) {
...
+ * Implements hook_preprocess_HOOK().
+ * for adding grid classes to multirow paragraphs
+ */
+function stanford_profile_styles_preprocess_react_paragraphs(&$variables) {
+  foreach ($variables['rows'] as &$row) {
+    foreach ($row['items'] as &$item) {
+      $width = $item['width'];
+      $item['attributes']->addClass("flex-$width-of-12");
+    }
+    $row['attributes']->addClass("container-$width-columns");
```
pookmish, line 331: now this will wrap the row with a class defined by the last item in the row. so if a row consists of `3, 3, 6` then the row will be wrapped with `container-6-columns`
If a row consists of `6, 3, 3` then the row will be wrapped with `container-3-columns`

src/scss/components/wysiwyg/_multi_per_row.scss
pookmish: sounds like an issue with decanter not respecting the class.

pookmish:
the row class still needs some work. and i would say all this work needs to be moved into `stanford_profile_styles`

stanford_basic.theme
jenbreese: Ok. Can we chat tomorrow about it so I see what you mean? Oh are you thinking ahead of when the author can change the number of columns? Instead of the space divided evenly?

pookmish: yep thats exactly what i'm thinking

jenbreese:
Closing this pull request since we moved the code to the Stanford Profile instead of here.

PR #152, opened 2020-04-10
D8CORE-1402: Fix for IE footer

cjwest:
# READY FOR REVIEW

# Summary
- Fix allowing you to be able to read footer items in IE11

# Review By (Date)
- Before code freeze

# Urgency
- Not

# Steps to Test

1. Look at footer in IE11
1. Check out this branch
1. Rebuild Cache
1. Look at the footer in IE11 again
1. Verify that you can read footer items.

Screen Shot 2020-04-09 at 4 58 01 PM

- D8Core, Maybe Decanter?

- JIRA ticket: https://stanfordits.atlassian.net/browse/D8CORE-1402
- Other PRs

This is a bit of a hack because it doesn't put footer items in the correct order. However because this is for IE11, I didn't know if I should dedicate any more time to this.

# See Also


src/scss/layout/_page.scss
```diff
@@ -17,3 +17,12 @@ body {
...
 }
+
+// Fix for IE11
+.su-local-footer__columns {
+  display: block;
+  @supports( grid-area: auto ) {
+    display: -ms-grid;
+    display: grid;
+  }
+}
```
sherakama, line 28: I would love this if it could be put into the Decanter project as that is where pretty near all the css for this component comes from: https://github.com/SU-SWS/decanter/blob/master/core/src/scss/components/local-footer/_local-footer.scss#L93

Also, this looks like a good opportunity to improve the `display-grid` mixin. https://decanter.stanford.edu/docs/scss/index.html#mixin-display-grid

sherakama: @cjwest

cjwest: @sherakama,
Thanks for looking at this.

See, the issue is that  IE11 doesn't support grid templates. However it _does_ support other grid functionality. And, when testing this out, I found that the fix needs to be fairly targeted. I did try overriding the {{display-grid}} mixin and found it was too general and messed up the whole site.  

I'll look into your first suggestion as that just targets just the grid templates.

sherakama: The autoprefixer is supposed to pick this up and fix it by creating `ms-grid-column` and `ms-grid-row` styles. 🤔

sherakama: I have had to fix that up due to the order in which they get rendered. See for example: https://github.com/SU-SWS/decanter/blob/master/examples/scss/layout/grid/three-column/_battleship.scss

cjwest:
It turns out that the autoprefixer in Webpack was missing a configuration item.
See https://github.com/SU-SWS/stanford_basic/pull/155

PR #159, opened 2020-05-11
D8CORE-2004 JS Error in stanford_basic local tabs.

rebeccahongsf:
# READY FOR REVIEW

# Summary
- Resolve JS error by checking for `role--authenticate` class.

# Review By (Date)
- asap (05/11)

# Urgency
- high

# Steps to Test
1. Pull this branch
2. Run `npm run publish`
3. Check out console log of local instance as anonymous user.

- Stanford_basic

- [D8CORE-2004](https://stanfordits.atlassian.net/browse/D8CORE-2004)

# See Also


src/js/theme/menu/StickyHeaderOnScroll.js
```diff
@@ -1,9 +1,11 @@
-window.onscroll = function () {
-  stickyHeaderOnScroll();
-};
-
-var header = document.getElementById('block-stanford-basic-local-tasks');
-var sticky = header.getBoundingClientRect().top;
+if (document.body.classList.contains('role--authenticated')) {
```
rebeccahongsf, line 1: Should `document.body.classList.contains('role--authenticated')` be declared as a variable or is this okay? 🤔 
(i.e. 
```
var authenticated = document.body.classList.contains('role--authenticated');
if (authenticated) { ... }
```
)

sherakama: How about simply checking if the block exists?

eg: 
```
var header = document.getElementById('block-stanford-basic-local-tasks');
if (!header.length) {
return;
}
```

rebeccahongsf: 🤦

src/js/theme/menu/StickyHeaderOnScroll.js
```diff
@@ -2,26 +2,28 @@ window.onscroll = function () {
   stickyHeaderOnScroll();
 };
 
-var header = document.getElementById('block-stanford-basic-local-tasks');
-var sticky = header.getBoundingClientRect().top;
-
 function stickyHeaderOnScroll() {
+  var authenticated = document.body.classList.contains('role--authenticated');
```
sherakama, line 6: It could be possible that someone is authenticated but has no local tasks block available. I would remove this check altogether and just validate that the block exists or not.

rebeccahongsf:
**Next steps:**
Resolve styling issue for News/People/Event pages where local tabs do not exist (see [JIRA ticket](https://stanfordits.atlassian.net/browse/D8CORE-2004))

src/js/theme/menu/LocalTaskToggle.js
```diff
@@ -0,0 +1,10 @@
+var localTab = document.getElementById('block-stanford-basic-local-tasks');
+var pageContent = document.getElementById('page-content');
+
+function localTaskToggle() {
+  if (localTab) {
+    pageContent.parentNode.classList.add('outline');
```
rebeccahongsf, line 6: Do we have a naming convention for when we add classes?
(I have the same question for the sticky class as well) 🤔

sherakama: We follow BEM: http://getbem.com/introduction/

For components we usually prefix them with `su-` but in this case you are not required to do that.

src/js/theme/menu/StickyHeaderOnScroll.js
```diff
@@ -1,9 +1,12 @@
-window.onscroll = function () {
-  stickyHeaderOnScroll();
-};
-
 var header = document.getElementById('block-stanford-basic-local-tasks');
-var sticky = header.getBoundingClientRect().top;
+var sticky = 0;
+
+if (header) {
+  sticky = header.getBoundingClientRect().top;
```
rebeccahongsf, line 5: @sherakama, something I noticed when testing locally — if a content author refreshes the page when they are half way down the page, the local tabs appear stick to the top (and upon refreshing this sticky distance is recalculated). So when scrolling to the top, the local tabs continues to stick to the header.

If the distance from the local tabs to the top of the page is consistent throughout the sites, should we use a fixed value instead of calculating the distance?

src/scss/theme/menu/_local-tasks.scss
```diff
@@ -13,7 +13,8 @@
 
     &:focus,
     &:hover {
-      border: 5px solid rgb(137, 206, 255);
+      outline: 5px solid rgb(137, 206, 255);
+      outline-offset: -5px;
       z-index: 103;
```
rebeccahongsf, line 18: Hmm although a `z-index` has been applied, the outline still appears behind full width items and the local tabs. 🤔

src/js/theme/menu/StickyHeaderOnScroll.js
sherakama: Chances are the value will change over time so it would be best to keep the calculation.

PR #160, opened 2020-05-12
outline instead of border

sherakama:
This is a suggestion to test out.

# READY FOR REVIEW 

# Summary
- Perhaps try out `outline` instead of border on the `page-content` region as that may help prevent the content shift that happens on hover.

# Review By (Date)
- meh

# Urgency
- low

# Steps to Test

1. Check out this branch
2. Compile the css (npm run publish) 
3. Clear cache and try it out on a page


pookmish:
The bounce is really annoying to me. Its mostly happening on full width pages that have a banner. And i think because of that bounce, if i hover over and away from the area enough times, my browser will scroll without me touching anything. its the craziest thing

sherakama:
FWIK outline is not only top and bottom but should be available on all four sides: https://www.w3schools.com/css/css_outline.asp

sherakama:
If the outline is disappearing because it is showing outside the window you should be able to adjust the offset. eg: `outline-offset: -10px`

rebeccahongsf:
> If the outline is disappearing because it is showing outside the window you should be able to adjust the offset. eg: `outline-offset: -5px`

 If we can eliminate the bouncing full width images by using `outline` with `outline-offset`, then this sounds good to me! 😄

sherakama:
Merge conflict resolved. Give it a go.

PR #165, opened 2020-06-05
D8CORE-000 Provide a default lockup option

pookmish:
# READY FOR REVIEW

# Summary
- With subthemes if the subtheme hasn't defined the lockup style option, it won't get appropriate styles and it will look wrong. This makes sure that there is always at least option `a` applied.

# Need Review By (Date)
- :shrug: 

# Urgency
- low

# Steps to Test
1. checkout the `8.x-4.x` branch.
1. `drush cdel stanford_basic.settings`
1. view the lockup and see how it doesn't look correct
1. checkout this branch and clear render caches
1. validate the lockup at least has the correct styles for option "a"

# See Also

### DIFF
--- a/stanford_basic.theme
+++ b/stanford_basic.theme
@@ -216,7 +216,7 @@ function _stanford_basic_preprocess_not_nodes(&$vars) {
  * Pass through the lockup configuration settings.
  */
 function stanford_basic_preprocess_block__system_branding_block(&$vars) {
-  $vars['lockup'] = theme_get_setting('lockup');
+  $vars['lockup'] = theme_get_setting('lockup') ?: 'a';
 }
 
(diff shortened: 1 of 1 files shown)


sherakama:
@boznik As you are working on providing a UI for lockup options does this affect what you are working on?

boznik:
Yes, this will affect the config page I am working on.

pookmish:
It shouldn't affect it. the config pages just does config overrides. `theme_get_setting()` just pulls from config. so if you have a config override, it'll come through still.

boznik:
ah ok, as long as it doesn't trump the override, i'm good with this.

sherakama:
Thanks all.

PR #166, opened 2020-06-08
D8CORE-1410: Back to site link

cjwest:
# READY FOR REVIEW

# Summary
- Placing a back to site button on the login page at /user

# Review By (Date)
when possible

# Urgency
Not urgent

# Steps to Test

1. Check out this branch
1. Navigate to /user/login
1. Back to site button should be on that page
1. Verify that link takes you to the homepage
1. Verify that it doesn't interfere with the Primary Tabs

- JIRA ticket: https://stanfordits.atlassian.net/browse/D8CORE-1410
- Other PRs

# See Also


cjwest:
@sherakama,

I could use some help with this:

1. I'm getting a 403 on the codeclimate page. Do I need different permissions? In any case, can you tell me what I need to fix?
1. I positioned the link absolutely so that it didn't interfere with the placement of anything else on the page. Let me know if I should have taken a different approach.
2. I wonder if the link should be part of page.content? Using voiceover, the primary navigation tabs jump around and hide the back to site button. This would also put it after the H1. I haven't figured out how I might add it to the page content though.

templates/page--user--login.html.twig
```diff
@@ -6,6 +6,12 @@
 
 <div class="page-content">
   <a name="page-content" id="page-content" tabindex="-1" class="visually-hidden focusable">Page Content</a>
+
+  <section class="su-back-to-site">
+    <h2><a href="/home" class="su-link su-button">
```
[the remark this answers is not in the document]
cjwest: Thank you!

cjwest:
Interesting. I pushed up some changes, and I can access codeclimate again. 🤷‍♀️

sherakama:
Screen Shot 2020-06-11 at 10 42 17 PM

This has some strangeness with the hover options. Actually, It would be good to resolve those hover to see tabs into static always up tabs or something for the login page.

sherakama:
@cjwest

cjwest:
@sherakama,
I don't see the tabs on this page on anything but my local. I figured if that was the case, I wasn't going to put more time into mucking with them. Are they new in this release and will be available on deployed sites?

sherakama:
I think it is fine to tackle fixing the positioning of the back-to-home button and the tabs in this ticket. It probably makes sense to disable the hover tabs on this page and go back to the regular tabs. @rebeccahongsf do you have a couple of cycles to assist?

rebeccahongsf:
@sherakama I’d love to take another pass on the local task tabs styling to resolve these issues!

rebeccahongsf:
Hey @sherakama, we were unable to reproduce the local tabs menu hover issue on production sites and my local. I believe the styling is only suppose to appear when the user is logged in. Perhaps this is something that only occurs on local when Drupal local login is enabled? 
I am wondering if we need to address this 🤔

sherakama:
We support this theme for more than just our ACSF product and we should support all configurations of the login page. To reproduce, stand up a local build and uninstall the simplesamlphp_auth and stanford_ssp modules. Visit the user page as an anonymous user.

templates/page--user--login.html.twig
```diff
@@ -6,7 +6,22 @@
...
+        <a href="{{ url('<front>') }}" class="su-link su-button">
+          <i class="fa fa-chevron-left"></i>
+          Back to your site
+        </a>
+      </div>
+    </div>
+    {{ page.content.stanford_basic_local_tasks }}
+  </div>
+
+  {{ page.content.stanford_basic_content }}
```
rebeccahongsf, line 23: Hi @sherakama! 😄 
Is this the best way to separate the `page.content` items like this? 
The reason why we wanted to do it this way was so that we can set the "Back to your site" button inline with the local task tabs.

sherakama: This, unfortunately, cannot be our path forward. If we went with this approach we would be unable to add or re-arrange any of the blocks in the block layout page. If you want to get the "back to your site" button inline with the local tabs block you will either have to adjust the `menu-local-task.html.twig` template and add conditionals or attach the markup to the block in a preprocess function instead.

rebeccahongsf: That makes sense! I’ll give this another pass. Thank you @sherakama 😃

sherakama:
@rebeccahongsf Is this up for review again?

rebeccahongsf:
@sherakama yes! Please let me know your thoughts. 😊

src/js/stanford_basic.behavior.js
```diff
@@ -33,6 +33,11 @@ export default {
         $('.page-content', context).addClass('stanford-basic--outline');
       }
 
+      var userLogin = $('.page-user-login', context);
+      if (userLogin.length === 0) {
+        $('.su-back-to-site', context).addClass('visually-hidden');
+      }
```
sherakama, line 39: This all looks great but I would like to see the reverse in logic. Have the back-to-site hidden by default (with css) and have it show when on the user login page. Just to be safe.

rebeccahongsf: The reverse logic has been implemented 😄  Let me know if any further changes are needed. Thank you @sherakama!

cjwest:
I mentioned this to Rebecca already. Just documenting this here:
I wonder if there's a better class than "visually-hidden"?
With "visually-hidden" is it still accessible by screen reader? If so, we might want a different class. Any suggestions?

sherakama:
> I mentioned this to Rebecca already. Just documenting this here:
> I wonder if there's a better class than "visually-hidden"?
> With "visually-hidden" is it still accessible by screen reader? If so, we might want a different class. Any suggestions?

yes, `hidden` should do it.

src/scss/theme/menu/_local-tasks.scss
```diff
@@ -23,7 +23,7 @@
 
 .block--local-tasks {
   @include margin(0 auto 10px auto);
-  z-index: 15;
+  z-index: 10;
```
rebeccahongsf, line 26: @sherakama may we include this small `z-index` change within this pull request as well?

This would resolve the local task bar overlapping issue with the dropdown navigation menu on mobile, in addition to the custom user-login dropdown links we have for the Cardinal Service profile.

PR #173, opened 2020-07-21
D8CORE-1333: Add '<main>' landmark

cjwest:
# READY FOR REVIEW

# Summary
- Adds `<main>` landmark

# Review By (Date)
- when possible

# Urgency
- Not critical

# Steps to Test

1. Checkout this branch
1. Clear the cache
1. Using WAVE, you can verify that there is now a `<main>` landmark

Screen Shot 2020-07-20 at 6 24 32 PM

- D8core

- JIRA ticket: https://stanfordits.atlassian.net/browse/D8CORE-1333
https://stanfordits.atlassian.net/browse/D8CORE-2444
- Other PRs

# See Also

### DIFF
--- a/templates/page.html.twig
+++ b/templates/page.html.twig
@@ -18,9 +18,9 @@
   </section>
 </header>
 
-<div class="page-content" id="page-content">
+<main class="page-content" id="page-content">
   {{ page.content }}
-</div>
+</main>
 
 <footer id="footer">
   {{ page.footer }}


sherakama:
Is `<main>` still appropriate if we have a sidebar or secondary/tertiary content within that content region?

sherakama:
@cjwest this seems reasonable, just one question about whether our layout templates that have more than one column can also be nested in main. The other option for adding a main tag would be to change the jumpstart_ui layouts but that could get trixy as they are used multiple times on a page quite often.

cjwest:
Good questions, @sherakama 

> The <main> element is used to denote the content of a webpage that relates to the central topic of that page or application. It should include content that is unique to that page and should not include content that is duplicated across multiple webpages, such as headers, footers, and primary navigation elements.

https://html.com/tags/main/

Note it doesn't say secondary navigation elements

sherakama:
Ok, so this is generally safe then. I am inclined to accept this and see if it turns up any accessibility errors.

cjwest:
@sherakama,

Let's hold.

We discussed this at SOAP office hours.  The issue was whether this makes sense due to the approach we're taking with the layouts. That is, with `<nav>` elements not at the top level, is it semantic to have this as `<main>`. My thinking that having the "Main" landmark is better than not, and that any secondary menu is going to be related to the current content. 

However, Sy, with Level Access wanted to consult with his colleagues.

cjwest:
I heard from Sy from Level Access:
It is valid from a mark up standpoint to have a <nav> inside a <main>
Let's go for it!

PR #186, opened 2020-09-23
Wrap page contents with attributes

pookmish:
# READY FOR REVIEW

# Summary
- Wrap the page contents in a div to allow attributes to be added
- pass `only` the brand bar variant to avoid inheritance of the rest of the attributes.

# Need Review By (Date)
- :shrug: 

# Urgency
- low

# Steps to Test
1. checkout this branch
1. create a `hook_preprocess_page` and add some attributes
1. see that the page has those attributes in a div.

# See Also

### DIFF
--- a/templates/page.html.twig
+++ b/templates/page.html.twig
@@ -1,28 +1,30 @@
-{%- include "@basic-dist/decanter/components/brand-bar/brand-bar.twig" with { modifier_class : brand_bar_variant } -%}
+{%- include "@basic-dist/decanter/components/brand-bar/brand-bar.twig" with { modifier_class : brand_bar_variant } only -%}
 
-{%- if page.help -%}
-  {{ page.help }}
-{%- endif -%}
-
-<header class="su-masthead su-masthead--right">
-  <section>
-  {%- if page.header -%}
-    {{ page.header }}
+<div{{ attributes }}>
+  {%- if page.help -%}
+    {{ page.help }}
   {%- endif -%}
-  {%- if page.search -%}
-    {{ page.search }}
-  {%- endif -%}
-  {%- if page.menu -%}
-    {{ page.menu }}
-  {%- endif -%}
-  </section>
-</header>
 
-<main class="page-content" id="page-content">
-  {{ page.content }}
-</main>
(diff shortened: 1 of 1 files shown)


templates/page.html.twig
```diff
@@ -1,28 +1,30 @@
...
-{%- endif -%}
-
-<header class="su-masthead su-masthead--right">
-  <section>
-  {%- if page.header -%}
-    {{ page.header }}
-  {%- endif -%}
-  {%- if page.search -%}
-    {{ page.search }}
+<div{{ attributes }}>
```
sherakama, line 3: YADW (Yet another div wrapper)

We are already in div soup. I would rather this go on the main element.

pookmish: but then its not the `page` attributes.

pookmish: I could've made it even better and wrapped the brand bar :laughing:

sherakama: I do like the `only` flag on the brand bar so thank you for that. 

What are you trying to do that requires preprocessing the page and attributes wrapping everything?

pookmish: I'm wrapping the page with a class when I'm on a node page and satisfy some conditions. I can do it in the CS theme so its not a blocker. But the `only` is kinda important

sherakama: You shouldn't be creating page-level CSS selectors. That creates strong css selectors that get difficult to override. Instead, use localized selectors and attach stylesheets in the appropriate places.

sherakama:
Adding @jenbreese @katrialesser to chime in and provide insight/strategy on implementation.

templates/page.html.twig
pookmish: Ooh. Yeah that could work for my situation. Conditionally add the styles heat :+1:

But that aside, I feel like if you can do a `hook_preprocess_hook` and add attributes, it should be reflected in the markup

sherakama:
Closing. As much as I agree with @pookmish that we should support conventional Drupal hooks in our theme, we don't want to support page wrapping and deeply nested selectors.

PR #190, opened 2020-10-09
D8CORE-2102: Add check to see if the path is part of the url

imonroe:
# READY FOR REVIEW

# Summary
- In the menu, the active item is set by checking if the URL is identical to the path of the menu item. Here, we add a check to see if the menu item path is part of the URL.

# Review By (Date)
- next week.

# Urgency
- 1 out of 10

# Steps to Test

1. Check out this branch
2. `drush cr`.
3. Create a news story
4. Visit `/news`.  Notice the "News" item in the main menu is active.
5. Click through to the news story you created.  Notice that the "News" in the main menu is still active.

- JIRA ticket
- Other PRs

# See Also

### DIFF
--- a/stanford_basic.theme
+++ b/stanford_basic.theme
@@ -250,9 +250,10 @@ function stanford_basic_preprocess_menu(&$variables, $hook) {
  */
 function _stanford_basic_menu_process_submenu(&$submenu, $current_path) {
   foreach ($submenu as &$item) {
-
     $item_path = (is_string($item['url'])) ? $item['url'] : $item['url']->toString();
-    if ($item_path == $current_path || strpos($current_path, $item_path . '/') !== FALSE) {
+    if ($item_path == $current_path ||
+        strpos($current_path, $item_path . '/') !== FALSE ||
+        ($item_path == '/home' && $current_path == '/')) {
       $item['is_active'] = TRUE;
     }
 
--- a/stanford_basic.theme
+++ b/stanford_basic.theme
@@ -253,7 +253,7 @@ function _stanford_basic_menu_process_submenu(&$submenu, $current_path) {
     $item_path = (is_string($item['url'])) ? $item['url'] : $item['url']->toString();
     if ($item_path == $current_path ||
         strpos($current_path, $item_path . '/') !== FALSE ||
-        ($item_path == '/home' && $current_path == '/')) {
+        ($item['title'] == 'Home' && \Drupal::service('path.matcher')->isFrontPage())) {
       $item['is_active'] = TRUE;
     }
 
--- a/stanford_basic.theme
+++ b/stanford_basic.theme
@@ -289,7 +289,7 @@ function _stanford_basic_path_is_home(string $path) {
       $normal_path == '<front>') {
         return TRUE;
   }
-  return false;
+  return FALSE;
 }
 
 /**
--- a/stanford_basic.theme
+++ b/stanford_basic.theme
@@ -249,7 +249,6 @@ function stanford_basic_preprocess_menu(&$variables, $hook) {
  * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
  */
 function _stanford_basic_menu_process_submenu(&$submenu, $current_path) {
-  $path_is_home = _stanford_basic_path_is_home('/');
   foreach ($submenu as &$item) {
     $item_path = (is_string($item['url'])) ? $item['url'] : $item['url']->toString();
     if ($item_path == $current_path ||
--- a/stanford_basic.theme
+++ b/stanford_basic.theme
@@ -252,7 +252,7 @@ function _stanford_basic_menu_process_submenu(&$submenu, $current_path) {
   foreach ($submenu as &$item) {
 
     $item_path = (is_string($item['url'])) ? $item['url'] : $item['url']->toString();
-    if ($item_path == $current_path) {
+    if ($item_path == $current_path || strpos($current_path, $item_path . '/') !== FALSE) {
       $item['is_active'] = TRUE;
     }
 
--- a/stanford_basic.theme
+++ b/stanford_basic.theme
@@ -249,11 +249,12 @@ function stanford_basic_preprocess_menu(&$variables, $hook) {
  * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
  */
 function _stanford_basic_menu_process_submenu(&$submenu, $current_path) {
+  $path_is_home = _stanford_basic_path_is_home('/');
   foreach ($submenu as &$item) {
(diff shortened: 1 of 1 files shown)


sherakama (dismissed):
OK by me.

jenbreese (changes requested):
Hi @imonroe, I noticed it isn't working on the mobile view. Thanks! Jen

jenbreese:
I did more testing. The news/article example works. On desktop view the Home link, the indicator works. On the mobile/responsive menu, the indicator is not there. 
Screen Shot 2020-10-16 at 8 10 15 AM
Screen Shot 2020-10-16 at 8 09 53 AM
Screen Shot 2020-10-16 at 8 09 43 AM

imonroe:
Ok, I fixed the issue with the Home link on mobile.  However, it's easy to see how this can lead to infinite regress if we're not careful with how we apply the `is_active` flag. E.g., the `People` menu item correctly is set to active when you're on the `/people` page, however, if you drill down to an individual, you'll find the menu path is `/person/john-doe` or whatever, and the path matching breaks.  We could build a lookup array with the basic menu items, but that's going to break as soon as folks start manipulating the contents of the menu on their individual sites.

stanford_basic.theme
```diff
@@ -250,9 +250,10 @@ function stanford_basic_preprocess_menu(&$variables, $hook) {
  */
 function _stanford_basic_menu_process_submenu(&$submenu, $current_path) {
   foreach ($submenu as &$item) {
-
     $item_path = (is_string($item['url'])) ? $item['url'] : $item['url']->toString();
-    if ($item_path == $current_path) {
+    if ($item_path == $current_path ||
+        strpos($current_path, $item_path . '/') !== FALSE ||
+        ($item_path == '/home' && $current_path == '/')) {
```
sherakama, line 256: I don't think we can assume home here. I've seen other paths used as well. You likely will have to get the alias from the site settings instead of assuming `/home`

imonroe: @sherakama what about something like `\Drupal::service('path.matcher')->isFrontPage()`?

sherakama: That might do it.

imonroe: Adjusted to allow for alternate front pages.

jenbreese:
@imonroe That fixed the highlight for "Home" on my dev for mobile. It is still working as expected for the desktop size.

stanford_basic.theme
```diff
@@ -250,9 +250,10 @@ function stanford_basic_preprocess_menu(&$variables, $hook) {
  */
 function _stanford_basic_menu_process_submenu(&$submenu, $current_path) {
   foreach ($submenu as &$item) {
-
     $item_path = (is_string($item['url'])) ? $item['url'] : $item['url']->toString();
-    if ($item_path == $current_path) {
+    if ($item_path == $current_path ||
+        strpos($current_path, $item_path . '/') !== FALSE ||
+        ($item['title'] == 'Home' && \Drupal::service('path.matcher')->isFrontPage())) {
```
sherakama, line 256: Do we need to check the title as well?

imonroe: @sherakama I'm not sure I understand exactly what you mean.  What we're checking here is that the title of the menu item is "Home", and the page we are actually on is set as the current front page.  We can check the title of the current front page to make sure it is what we expect, but that seems redundant, yeah?  The title of the current front page is somewhat arbitrary, since the site owner can set it to any page they like...

sherakama: As the title for the link can be arbitrary, I was wondering if we want to check it at all? Shouldn't this function just try to match the URL, in whatever form it is in, to the current page/URL? It does appear the home page needs special handling as it could be `/home` `/` or perhaps even `<front>`

eg:
```
if (...->isFrontPage()) {
  $ops = ['/', '/home-page-alias', '/node/12345', '<front>'];
  if (in_array($item_path, $ops)) {
    $item['is_active'] = TRUE;
  }
}
```

imonroe: I believe that the result of that code will be that if you are on the front page of the site, every link in the navigation menu will be marked as active.  

So, we're looking at `$item['title']` to know if we are manipulating the "Home" link in the menu.  The `$item_path` variable will be `/` for that, which we could target as well I suppose, which would allow the user to change the name of the "Home" link in the menu.  

To put it another way, if you are on the homepage of the site, regardless of what page that is, and you haven't changed anything in the menu, then: 
`$item['title']` == 'Home'` OR `$item_path == '/'` will both target the same button in the navigation menu, and `...->isFrontPage()` will return TRUE.

However, if the site owner has changed the link title in the menu, or has changed the link path in the menu, then one or both of those targets will no longer be useable.  

My thinking was that by targeting the button based on `$item['title']` == 'Home'` was slightly more robust, since a site owner who has changed their front page may have changed the link (the `$item_path`) to the page they want as the homepage, while leaving the text of the link (`Home`) alone.

sherakama: I think we can ignore the text of the link for this check altogether. It is likely to be 'home' but not necessarily the case as you outlined above. From what I know, the home page could be a number of valid URLs and we will need to check for each type. Off the top of my head a few options are:

- `/` Root of the domain
- `/[base_path]` Where this site is configured to sit under a path under a TLD. eg: https://example.com/mysite/
- `/[alias]` Where alias is the node path alias
- `/node/[nid]` Path to the direct node
- `<front>` Special keyword
- `https://[vanity]'.stanford.edu` if the user does this they don't deserve highlighting.

I would rather see a check for those combinations of URLs instead of checking for a text label.

stanford_basic.theme
```diff
@@ -266,6 +267,31 @@ function _stanford_basic_menu_process_submenu(&$submenu, $current_path) {
...
+  // account for weird paths in input
+  $normal_path = strtolower(trim($path));
+  $config = \Drupal::config('system.site');
+  $front_uri = $config->get('page.front');
+  $alias = \Drupal::service('path.alias_manager')->getAliasByPath($front_uri);
+
+  if ($normal_path == base_path() ||
+      $normal_path == $front_uri ||
+      $normal_path == $alias ||
+      $normal_path == '<front>') {
```
imonroe, line 289: `base_path()` accounts for `/` or `/base/path/`
`$front_uri` accounts for `node/10`, etc.
`$alias` accounts for `/home`, etc.
`<front>` accounts for front, obvs.

stanford_basic.theme
imonroe: @sherakama see notes on the update below.

stanford_basic.theme
sherakama: This was a pain but thank you for doing it. This is much more robust.

sherakama (approved):
GTG

PR #198, opened 2020-11-11
D9 Readiness

pookmish:
# READY FOR REVIEW

# Summary
- replaced deprecated service
- Moved preprocess into stanford_profile_helper since it needs to be part of a class in D9 https://github.com/SU-SWS/stanford_profile_helper/pull/50

# Need Review By (Date)
- End of the year

# Urgency
- low

# Steps to Test

# See Also

### DIFF
--- a/stanford_basic.theme
+++ b/stanford_basic.theme
@@ -318,37 +318,6 @@ function _stanford_basic_link_is_public(Url $url) {
 }
 
 /**
- * Implements hook_preprocess_ds_entity_view().
- */
-function stanford_basic_preprocess_ds_entity_view(&$variables) {
-  $variables['content']['#pre_render'][] = 'stanford_basic_pattern_process_contextual';
-}
-
-/**
- * PreRender process to render out the contextual links for ui patterns.
- *
- * @param array $element
- *   Pre-rendered element.
- *
- * @return array
- *   Modified element.
- *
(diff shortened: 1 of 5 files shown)


stanford_basic.theme
```diff
@@ -317,37 +317,6 @@ function _stanford_basic_link_is_public(Url $url) {
...
-  $module_handler = \Drupal::moduleHandler();
-  if (isset($element['#contextual_links']) && $module_handler->moduleExists('contextual')) {
-    $placeholder = [
-      '#type' => 'contextual_links_placeholder',
-      '#id' => _contextual_links_to_id($element['#contextual_links']),
-    ];
-    $element['#prefix'] = render($placeholder);
-  }
-  return $element;
-}
```
sherakama, line 349: As all of this is all going away and we don't know who is using this base theme you should add a changelog entry, deprecation notice, and document upgrade path to restore the behavior.

pookmish: How do you suggest we add a deprecation notice? normally we'd do that and then do a release indicating that the deprecation will be removed in the next Minor release instead of patch release.

pookmish: I added a part to the changelog and an `UPGRADE.md` file for more information on how to restore it.

sherakama (approved):
GTG

PR #206, opened 2021-01-14
D8CORE-1123: Enable dropdowns for desktop

imonroe:
# READY FOR REVIEW

# Summary
- This PR adds support for dropdown navigation menus at desktop resolutions for stanford_basic

# Review By (Date)
- end of sprint

# Urgency
- it's a feature for the 2.0 release.

# Steps to Test

1. Checkout this branch.
2. Clear your caches.
3. Add a page or two, and add menu links for them.
4. go to `/admin/structure/menu/manage/main` and set up some of your pages so they are sub-items of other menu itmes.
5. go to `/admin/appearance/settings/stanford_basic` and check the option for "Enable dropdowns for navigation menu"
6. Save your changes, clear your caches.  Visit any page and check to make sure the drop down menus work for you.

- Acceptance tests in the profile here: https://github.com/SU-SWS/stanford_profile/pull/349

# See Also


stanford_basic.theme
```diff
@@ -23,6 +23,8 @@ function stanford_basic_page_attachments_alter(array &$attachments) {
   if ($route_match->getRouteName() == 'user.login') {
     $attachments['#attached']['library'][] = 'stanford_basic/user_login';
   }
+  // Check if dropdown menus are activated
+  $attachments['#attached']['drupalSettings']['nav_dropdown_enabled'] = (theme_get_setting('nav_dropdown_enabled')) ? TRUE : FALSE;
```
pookmish, line 27: This should be namespaced and then use that namespacing in the javascript.
```suggestion
  $attachments['#attached']['drupalSettings']['stanford_basic']['nav_dropdown_enabled'] = (theme_get_setting('nav_dropdown_enabled')) ? TRUE : FALSE;
```

src/js/components/secondary-nav/buttons/SecondarySubNavButtons.js
```diff
@@ -58,6 +58,17 @@ export default class SecondarySubNavButtons {
 
     // Add the accessibility meta-information.
     this.initAccessibility();
+
+    if (window.innerWidth >= 992 && !drupalSettings.nav_dropdown_enabled) {
+      // We are on desktop, and dropdowns are disabled.
+      // Remove the toggle button.
+      this.item.getElementsByTagName('button')[0].remove();
```
pookmish, line 65: This might be for later, what happens when you change the window widths. Like going from landscape to portrait mode.

src/js/components/multi-menu/common/MobileToggle.js
```diff
@@ -19,7 +19,7 @@ export default class MobileToggle {
     this.elem = element;
     this.nav = nav;
 
-    // Merge options with defualts.
+    // Merge options with defaults.
```
pookmish, line 22: this is good difults

pookmish:
I'll test the functionality tomorrow. I don't see anything glaring at me, but i'll lean on @jenbreese and/or @sherakama  for scss and js review

src/js/components/secondary-nav/buttons/SecondarySubNavButtons.js
imonroe: When the screen drops below 992px, the hamburger menu takes over.  I've tested changing resolutions, and it behaves like I would expect it to.

pookmish:
There appears to be an issue with the last menu item. see the screenshot 
image

imonroe:
> There appears to be an issue with the last menu item. see the screenshot

Can confirm this is an issue.  Working on a fix.

sherakama:
This is a large enough PR that I won't be able to get to it today. When do you need it by? Not that my review should block this as you have other outstanding developers providing their review.

stanford_basic.theme
imonroe: I fixed this.

src/js/components/secondary-nav/buttons/SecondarySubNavButtons.js
imonroe: So actually, this was a problem in some cases.  I refactored to remove this JS, and replaced it with CSS using a modifier class to control the visibility of those elements.

src/js/components/multi-menu/common/MobileToggle.js
imonroe: it's the little things.  :)

imonroe:
> This is a large enough PR that I won't be able to get to it today. When do you need it by? Not that my review should block this as you have other outstanding developers providing their review.

@sherakama Early next week.  I still need to hand this off to @cjwest for a11y and then to Jacob for dox, prior to 2.0 release.

jenbreese:
Hi @imonroe, I noticed the lines separating the links in the desktop view don't line up on the ends. There's also a bottom border on the last menu item that might need to go. Or has design reviewed it too? 
Screen Shot 2021-01-23 at 8 48 17 AM

Screen Shot 2021-01-23 at 9 00 22 AM

Screen Shot 2021-01-23 at 9 01 17 AM

I can't use the keyboard to tab to the arrow and get focus to the menu arrow.

I tried it on Safari, FF and Chrome. I will look at the code more closely next. This is what I noticed on the menu display.

imonroe:
@jenbreese Thank you for finding the margin problem, it had escaped my notice.  Update coming shortly.

pookmish:
So there's only one visual issue i found. 
image

But there is some usability issues that may or may not be part of this. In the screenshot, there is a non-linking item. It still produces a link with an empty `href`. I assume Decanter doesn't have logic for that scenario.

The other issue is the checkbox on the menu item "Show as expanded". I assume this is overridden on the block to display them for mobile displays. I wonder if we can accommodate that. Just a random thought would be to preprocess the menu link item and add a data attribute like `data-hide-children` and css would force the child menu to be hidden on desktop. That could be accomplished in a later release though.

src/js/components/multi-menu/buttons/events/OnArrowRightLV1.js
```diff
@@ -1,29 +1,23 @@
...
+   * Handle the events for desktop sized screens.
    */
-  exec() {
-    if (this.isDesktop()) {
-      this.handleDesktop();
-    }
-    else {
+  handleDesktop() {
+
+    if (drupalSettings.stanford_basic.nav_dropdown_enabled) {
```
sherakama, line 16: What happens when you sub-theme? Does this setting change to the name of the theme? I think it does, you may want to make this more variable.

src/js/components/multi-menu/buttons/events/OnArrowUpToggleLV1.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Event action handler class.
+ */
+export default class OnArrowUpToggleLV1 extends MultiMenuEventAbstract {
+
+  /**
+   * Handle the events for desktop sized screens.
+   */
+  handleDesktop() {
+    if (drupalSettings.stanford_basic.nav_dropdown_enabled) {
```
sherakama, line 15: Same here. `drupalSettings.stanford_basic` could change on sub-themes.

sherakama: Ah, I see you add it specifically in the .theme file. It would be good to test that this still works on sub-themes.

imonroe:
I fixed the visual quirks that @jenbreese noticed; the menus line up on the left now, as per the designs.  I also fixed the keyboard navigation to work with the arrow keys correctly.  Tested with both the dropdowns enabled and disabled.  

@pookmish -- the different color you're seeing isn't the active menu item, it's a menu item in the active trail.  As for non-linking items, that wasn't something we'd covered, and it's one of the drawbacks of using the dropdowns in the first place.  I think that's an issue to cover in the docs, but I'm not sure what we'd do about it besides remove the <a> tag.  That would mess with styling though, since the <a> tag is what holds the actual menu label.  

@sherakama -- I added the `stanford_basic` namespace to that variable at the request of @pookmish.  I think it's something we'll have to change for subthemes, but let me see if I can find a way to automate.

src/js/components/multi-menu/buttons/events/OnArrowRightLV1.js
imonroe: refactored to make the theme name namespace a variable.

src/js/components/multi-menu/buttons/events/OnArrowUpToggleLV1.js
imonroe: refactored to make the theme name namespace a variable.

imonroe:
Per @ishahism, I added a separator bar to the top-level menu items with dropdown toggles:

image

sherakama:
> @sherakama -- I added the `stanford_basic` namespace to that variable at the request of @pookmish. I think it's something we'll have to change for subthemes, but let me see if I can find a way to automate.

As long as the variable is still present and the drop-down still works when a sub-theme is active, having the name be `stanford_basic` is ok. Just wanted to check that it does translate down.

ishahism:
Hi @imonroe, thank you very much for incorporating the horizontal divider and pushing pixels with me to provide a more supportive UI. I wanted to post our conversation about **how can we create more balanced spacing between "Resources 
  | v" and "News"...  how can we increase the spacing between the "drop-down caret" and "News" link so it has the same separation as "News" and "People"?**  _*I know you were looking through code to see if there is a way. But maybe @sherakama or @yvonnetangsu might have some additional thoughts since they worked on Decanter?_

ishahism:
Also @imonroe, I do want to acknowledge that you have created the interaction we originally discussed some 2 months back. I was just under the impress that Decanter had the horizontal line in the works before their work was paused. :-/

ishahism:
Oh and wanted to clarify the color difference for active state. @pookmish, the different color active state on the main navigation is intentional:

- Resources will have a black bar if the user is on the Resource landing page
- Resources will have a "tan" bar if the user is on any of the sub-pages nested under Resources

yvonnetangsu:
> Hi @imonroe, thank you very much for incorporating the horizontal divider and pushing pixels with me to provide a more supportive UI. I wanted to post our conversation about **how can we create more balanced spacing between "Resources | v" and "News"... how can we increase the spacing between the "drop-down caret" and "News" link so it has the same separation as "News" and "People"?** _*I know you were looking through code to see if there is a way. But maybe @sherakama or @yvonnetangsu might have some additional thoughts since they worked on Decanter?_
> 
> Here is a mock-up of the proposed design:
> Split button Drop-Down
> 
> Please let me know if I can provide more clarification.

Hi @ishahism ! Is this code deployed somewhere on a test site? I'd be happy to take a look if I could inspect it in the browser - my Stanford basic local repo is unfortunately not up to date so would take quite a while if I need to pull this down locally.

Also, FYI I see a MS(-3) in the mockup - heads up MS(-3) is going away since modular-spacing is being replaced by reponsive-spacing (which does not have a -3 or -4 step). We decided that for spacing this small, there is no need to write 3 separate media query for it (which is what modular/responsive spacing does). Using `em` as the unit is an alternative :)

ishahism:
Hi @yvonnetangsu, I don't believe this code has been deployed. Ian and I had our conversation while reviewing things on his local machine.

yvonnetangsu:
> Hi @yvonnetangsu, I don't believe this code has been deployed. Ian and I had our conversation while reviewing things on his local machine. (Thank you for the heads up on the "negative" spacers :)

Hi @ishahism - Completely taking a guess here since I can't inspect in the browser. It might be possible to add `margin-right` to the carets (looks like they are ::after pseudo elements, though the behavior also depends on how some other CSS properties are set). It might also have to do with where the spacing is applied - whether it's in the list item `li` or in the link `a` itself :)

ishahism:
Hi @imonroe, sorry for mudding the conversation here with design notes! I want to clarify that the functionality we discussed yesterday works as intended. My comments are about the visual spacing that will better support usability.

imonroe:
Following our conversation yesterday, I made the visual updates to the top-level menu to match the pixel measurements given above for the spacing and the separator bar.

While I was at it, I went through and simplified the SASS to get everything related to the multi-menu component into the same file.  That should make it substantially easier to make changes and updates going forward.  It would be super if @jenbreese could have a look at the refactored `_multi-menu.scss` to confirm I haven't done anything monsterously wrong, but I think it's in pretty good shape.  I tested it at a variety of resolutions, for mobile, for desktop, with and without dropdowns enabled, and it looks pretty good to me.

image

pookmish (dismissed):
This looks great. lets put a bow on it at this time and get some more eyes on it on the test sites.

jenbreese:
Hey @imonroe I will pull in your changes and take a look.  thanks!

jenbreese:
I am getting this error in FF and Safari. Has there been more changes I need besides the styles?  
Screen Shot 2021-01-28 at 12 01 37 PM

imonroe:
I fixed the FF javascript bug.  I was too clever by half.

jenbreese:
> I fixed the FF javascript bug. I was too clever by half.

I confirm that is fixed on my local for both FF and Safari.

jenbreese:
I tested it out and it is looking good to me. I didn't find any issues on FF and Safari.  I think it should be merged in so folks can try it out

imonroe:
Mike gave thumbs up so I'm admin-merging.

PR #207, opened 2021-02-01
Adjust `theme_get_setting` to target this theme, so that subthemes inherit it.

pookmish:
# READY FOR REVIEW

# Summary
- Allow children to inherit the dropdown settings.
- do we want to do this or require the subthemes to do their own settings?

# Need Review By (Date)
- today

# Urgency
- medium

# Steps to Test

# See Also


imonroe:
So I tried doing this by detecting the theme name from the `drupalSettings` object in JS, but that approach ended up breaking with FF and Safari, though it worked in Chrome.  I'm good with this solution.

imonroe (approved):
Ok with me.

PR #218, opened 2021-03-10
D8CORE-3474: unsetting the white color to fix for a11y

jenbreese:
# READY FOR REVIEW

# Summary
- Unset the white color on the search icon/button
- Fixed SOE and SBT

# Review By (Date)
- 3/12

# Urgency
- High 
- A11y violation

# Steps to Test

1. Pull in this change and clear caches.
2. Notice no visual change to the search icon
3. Run Wave.
4. Notice no more error there.

- Search button
- SOE
- SBT
- SAA? @katrialesser 

- D8CORE-3474
- @katrialesser I did a preliminary test and there was no visual change to SAA. It should fix it's a11y contrast error too.

# See Also


src/scss/components/masthead/_wrapper.scss
```diff
@@ -84,6 +84,7 @@
   }
 
   .su-site-search__submit {
+    color: unset;
     opacity: 100;
     right: 1.6rem;
   }
```
cjwest, line 90: Hi @jenbreese,

I'm curious if you tried the mixin `hide-visually`
It's used here:
https://github.com/SU-SWS/decanter/blob/master/core/src/scss/core/helpers/_sr-only-text.scss
or `hide-text`
https://github.com/SU-SWS/decanter/blob/master/core/src/scss/core/helpers/_sr-only-text.scss

jenbreese: Hi @cjwest, I did it that way because the button already has the class.  The white color doesn't actually show anywhere and I thought Wave was catching it as a false positive. 
Screen Shot 2021-03-12 at 10 25 57 AM

jenbreese: Hi @cjwest, I did it that way because the button already has the class.  The white color doesn't actually show anywhere and I thought Wave was catching it as a false positive. I'm doing some testing.  Oo, I see the class names are super similar but different. More to come. :)

jenbreese: Hi @cjwest, It already has the `@mixin hide-text` and I tried the `@include sr-only;` which removed the icon entirely. It is inheriting the button color here: https://github.com/SU-SWS/decanter/blob/master/core/src/scss/utilities/mixins/button/_button-primary.scss

The icon doesn't need a color so I think unsetting makes the most sense to me.

imonroe (approved):
+2/-1?  How can I say no?

PR #222, opened 2021-03-22
D8CORE-3682: Fixed word wrapping on long menu titles

imonroe:
# READY FOR REVIEW/

# Summary
- Fixes word wrapping for very long menu items.

# Review By (Date)
- 🤷 

# Urgency
- low urgency

# Steps to Test

1. Checkout this branch locally.
2. clear your caches.
3. Create a basic page with a really long title, and give it a menu entry.  Make it a sub-menu of another menu entry.
4. Activate dropdown menus.
5. Check the dropdown to make sure the very long item wraps gracefully.

- D8CORE-3682
- Other PRs

# See Also


src/scss/components/multi-menu/_multi-menu.scss
```diff
@@ -280,6 +280,8 @@
 
           @include padding(null 0 2.4rem 1rem);
           @include margin(0 1.8em 0 0.5rem);
+          max-width:90%;
+          word-wrap:break-word;
```
pookmish, line 284: Why 90%?
```suggestion
          max-width: 90%;
          word-wrap: break-word;
```

imonroe: Because I needed a relative measurement that wouldn't affect the ::after pseudo-selector positioning too much.  I am open to argument if there's a better unit or value : 
image

pookmish: Ahh. Would a `calc(100% - 20px)` be better then? Just a thought...

imonroe: It appears to make no difference, six of one, half-dozen of the other.

image

pookmish: Yeah it probably only would make a difference if the container was allowed to stretch far. like 90% of a 1200px container is 1080px and 100% - 20px is 1180px so it's a big difference. but on something of this size its probably a difference of a few pixels that isn't even noticeable because of the text size.

pookmish (approved):

PR #226, opened 2021-04-29
D8CORE-4037: fixed word wrapping with very long menu items

imonroe:
# READY FOR REVIEW

# Summary
- In some, but not all, cases, long menu items weren't wrapping correctly when there was a submenu for the item.  This corrects that problem by adjusting the max-width for items that have a submenu below them, and removes the unnecessary width restriction when there is no submenu.

# Review By (Date)
- end of sprint
- 
# Urgency
- low 

# Steps to Test

1. check out branch, clear caches.
2. create some pages that have long titles, and set them up as drop down menus.
3. make several levels if you can.
4. verify that the long menu items wrap correctly both with submenus and without.

- JIRA ticket
- Other PRs

# See Also


pookmish (changes requested):
The approach looks fine. but theres a couple issues if we can fix them:
The word is really close to the right edge.
image

Also if a page is unpublished, the indicator is mostly behind the menu item above it. probably just needs a higher z-index.

imonroe:
Looks like the z-index thing is an existing bug.  just setting the z-index doesn't fix it, because it's got CSS animation on it, and that re-sets the z-index to 0 when it redraws.  I'm trying to see if I can find a fix, but it's trickier than it seemed.

imonroe:
I fixed the margin for items that have no sub menu, but I'm lost on the stacking thing.  Because the opacity is adjusted as part of the css animation, it always resets the z-index to `auto`, and I don't know how to make it not do that, since it's in the css specs.  @jenbreese do you have an idea?  example:

image

jenbreese:
@imonroe  Try adding this on _menu.scss at link line
`.su-multi-menu__item--expanded {
  .unpublished-indicator {
    top: 0;
  }
}`

imonroe:
That did the trick!  Thanks very much Jen!

src/scss/admin/_menu.scss
```diff
@@ -16,6 +16,8 @@ a:focus {
   }
 }
 
+.su-multi-menu__item--expanded { .unpublished-indicator { top: 0; } }
```
pookmish, line 19: a couple things here
1. move this to the [admin scss](https://github.com/SU-SWS/stanford_basic/blob/8.x-5.x/src/scss/admin/_menu.scss#L19)
2. this doesn't look good as it overlays on top of the link text. This can easily be fixed by removing the [`overflow: hidden](https://github.com/SU-SWS/stanford_basic/blob/8.x-5.x/src/scss/components/multi-menu/_multi-menu.scss#L277) and positions the indicator in the normal position

pookmish (changes requested):

pookmish:
Still an issue with the spacing 
image

src/scss/admin/_menu.scss
jenbreese: I'm curious, do you mean the 'overflow: hidden' on the parent 'li'? In my test, when I remove that it changes the width of the indictor. I don't know if it was intentional to narrow that indicator?

imonroe:
> Still an issue with the spacing
> image

Again, I think you may not have the newest version of the CSS pulled in, or maybe something is cached.  I can't reproduce with that same title: 

image

imonroe:
I can pull that max-width in a few more pixels; it might be differences in the way your linux browser is drawing the fonts, etc, because I cut that edge off as close as I could to the drop-down toggle to make it as wide as possible.

src/scss/admin/_menu.scss
imonroe: Actually, removing overflow:hidden there obviates the need for the additional `top: 0;` all together.  I see what Jen means about the active indicator getting a pixel or two wider, but I think it actually looks a little better that way.  It makes it the same width as the active trail indicator.  I've made the change.

imonroe:
I reduced the width of the text-wrapper and I removed `overflow:hidden` from the `li` element.  Should be good now.

pookmish (approved):
Ok this looks good to me now :+1:

PR #227, opened 2021-05-04
D8CORE-3502 - a11y fixes for nav menu

imonroe:
# READY FOR REVIEW

# Summary
- This PR addresses a number of a11y concerns related to the main nav menu.

# Review By (Date)
- 5/6/2021

# Urgency
- medium

# Steps to Test

1. Checkout this branch.  Clear your caches, including browser cache.
2. Enable drop-down menus, create a bunch of menu items, and nest them however you want.
3. View the nav menu in the front end. verify:
	- `aria-expanded` no longer appears on `<a>` tags, but DOES appear on the drop-down toggles.
	- the `aria-label` on the drop down toggles no longer says 'expand menu', but rather mentions the name of the menu the toggle opens.
	- the `esc` key returns focus to the drop-down toggle when used to close a menu, rather than to the parent menu item.
	- focus indicators of both color and underline are present on the drop-down toggles in both firefox and edge, if possible.  I tested this myself, and I found they were present, so I didn't change anything about them, but it was noted as a bug in the ticket.

- JIRA ticket
- Other PRs

# See Also


jenbreese:
@imonroe Does this PR have the long title fix in it?

jenbreese:
@imonroe here you said `the aria-label on the drop down toggles no longer says "expand menu", but rather mentions the name of the menu the toggle opens.`. On my local, it still says "expand menu"

imonroe:
> @imonroe here you said `the aria-label on the drop down toggles no longer says "expand menu", but rather mentions the name of the menu the toggle opens.`. On my local, it still says "expand menu"

Make sure you clear your browser cache.  In Chrome, open the dev tools, and then right click the "reload" button in the address bar.  that'll bring up an option to clear caches and hard reload.

imonroe:
> @imonroe Does this PR have the long title fix in it?

That one is in the other PR. https://github.com/SU-SWS/stanford_basic/pull/226

jenbreese:
Hi @imonroe I tried a couple times clearing the caches and reloading and clearing the browser cache. I still am getting `expand menu`.
Screen Shot 2021-05-04 at 1 48 40 PM

src/js/components/secondary-nav/buttons/SubNavToggle.js
```diff
@@ -32,6 +32,11 @@ export default class SubNavToggle {
     // Assign the event dispatcher and event registry.
     this.eventRegistry = this.createEventRegistry(options);
     this.dispatch = new EventHandlerDispatch(element, this);
+
+    // Label the toggle to indicate what menu it opens
```
imonroe, line 36: @jenbreese This was the part that was missing.  It was still in my undo buffer, so a CTRL-Z or three got it back for me.

jenbreese: @imonroe So I was testing tonight. The aria-label on the drop down toggle worked on Mac/Chrome. All the other aspects worked too.  For Firefox,  the aria-label does not work on Mac, or on Windows (browserstack). The escape key does work and the aria-expanded is on the right element.  I tried it on Edge and Windows (browserstack and everything worked.  Looks like FF is the problem for that label. :)

imonroe: I found the bug and fixed it.  I've gotta remember to use FF only when writing JS; it's so picky.

pookmish (changes requested):
This doesn't appear to work correctly. It concats  all the parent menu items togeather for child button aria labels 
image

imonroe:
> This doesn't appear to work correctly. It concats all the parent menu items togeather for child button aria labels
> image

I believe you don't have the latest version, or you didn't clear your caches.  That was happening in an earlier version of the code, but I fixed it.

image

pookmish:
It appears pulling the latest commits fixed some of it, but I don't believe there's a need to concat the parent menu item to the current one. "Open the [menu item] menu" not "Open the [parent item]+[menu item] menu" makes very long read out for the users when you get down to the 3rd and 4th levels.

imonroe:
@pookmish I found a simpler way of grabbing the link text from the link, and it fixed the problem with the concatenated menu names.  Should be g2g now.

pookmish:
@imonroe ok that does work better. but the "Open" word doesn't change when the menu is open vs closed. The language should change depending if the submenu is open or not.

imonroe:
> @imonroe ok that does work better. but the "Open" word doesn't change when the menu is open vs closed. The language should change depending if the submenu is open or not.

How about if we just change it to "toggle" instead?  covers both situations.

pookmish:
> How about if we just change it to "toggle" instead? covers both situations.

The language needs to change. A visually impaired user won't know if the sub menu is open or not based on "Toggle"

imonroe:
> > How about if we just change it to "toggle" instead? covers both situations.
> 
> The language needs to change. A visually impaired user won't know if the sub menu is open or not based on "Toggle"

Ok, I added an Observer to watch for changes in the aria-expanded, and if it finds one, it triggers a label change based on the current state.  Works pretty well for me.  Lemme know if that's sufficient, @pookmish .

pookmish (approved):
This looks good to me now. Thanks for working through those fixes :+1:

src/js/components/secondary-nav/buttons/SubNavToggle.js
```diff
@@ -32,6 +32,20 @@ export default class SubNavToggle {
     // Assign the event dispatcher and event registry.
     this.eventRegistry = this.createEventRegistry(options);
     this.dispatch = new EventHandlerDispatch(element, this);
+
+    // Label the toggle to indicate which menu it opens
+    this.elem.setAttribute('aria-label', 'Open the ' + this.parentNav.elem.innerText.trim() + ' menu');
+
+    // Create an observer to watch if the toggled menu changes state.
+    var self = this;
+    this.observer = new MutationObserver(function () {
```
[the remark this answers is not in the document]
imonroe: it's vanilla JS, which is convenient.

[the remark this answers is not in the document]
sherakama: Cool: https://caniuse.com/mutationobserver

PR #229, opened 2021-05-12
D8CORE-2008: Removing the secondary skip link when the hamburger visible

jenbreese:
# READY FOR REVIEW

# Summary
- When you are on a page with secondary navigation and the mobile responsive hamburger menu is visible, remove the skip to secondary navigation link because it does nothing.
- **NOTE**: it doesn't respond to when you resize the browser window. @cjwest mentioned she thought it was ok since if you are using a screen reader you won't be resizing the window. Is this the best solution? Or should it be responsive? My thought was since it was for keyboard and screenreaders and not visible otherwise, this seemed better. The other option would be to track the window size changes.

# Review By (Date)
- 5/15

# Urgency
- med

# Steps to Test

1. Pull in this change.
2. Clear caches
3. Create a secondary navigation.
4. Shrink the browser so you have a hamburger menu like this:
5. Screen Shot 2021-05-18 at 7 53 37 AM
5.  Go to the page that has a secondary navigation menu on it.  Press tab to see only the `Skip to main content`. 
6. You should not see this:
7. Screen Shot 2021-05-18 at 7 57 57 AM
8. The idea is that if you have the hamburger menu you shouldn't get that `Skip to Secondary Menu` link because it is a useless link.
9. In the above note, I tell about the limitations of this.

- Stanford Basic theme

- D8CORE-2008
- @cjwest 

# See Also


pookmish:
@jenbreese can you please provide more testing tests. How do I test this PR?

jenbreese:
@pookmish Oh man, I'm sorry. I just updated it. Thank you!

src/js/stanford_basic.behavior.js
```diff
@@ -20,9 +20,10 @@ export default {
       }
 
       // Validate there is a skip link for the secondary navigation. If not,
-      // remove the skip link.
+      // remove the skip link. If the mobile hamburger is visible, remove the link.
       var $sn = $('#secondary-navigation', context).length;
-      if (!$sn) {
+      var $snm = ($('.su-multi-menu__nav-toggle').css('display') === 'flex' );
+      if (!$sn || $snm) {
```
pookmish, line 26: I don't believe we should actually `remove()` the link. What if a user expands the window or goes from portrait to landscape. The link is gone and no longer available.

jenbreese: @cjwest said that since this is only used by screenreaders it is unlikely they are going between portrait and landscape views. Caryl, can you confirm that?

cjwest: @jenbreese and @pookmish,
Mike makes a good point. I know, Let's just prohibit people from changing screen sizes!!!! 🎉 
Okay, seriously, I don't know. 
One thing would be to just use CSS to hide the link from screen readers for specific screen sizes. That way we just reveal it if the screen size changes. 
I'm happy to talk this through and to bring up with SODA. Your thoughts?

jenbreese: @pookmish This is what SOAP said we should do "Oh, they said to go ahead and hide the link secondary nav with display:none (or something from decanter) then reveal it if the screen size should change."

I'll fix this next.

PR #230, opened 2021-05-17
D8CORE-2494 Add "Back to Top" button

pookmish:
# READY FOR REVIEW

# Summary
- Add a "Back To Top" button
- Only displays if the scroll position is over 3 times the current window height.

# Need Review By (Date)
- 5/20

# Urgency
- low

# Steps to Test
1. checkout this branch and clear caches
2. scroll down on a long page
3. click the "Back to top" in the bottom right
4. see that it takes you to the page top.

# See Also


jenbreese:
I tried it on Browserstack (mobile devices), Firefox and Chrome and it all worked as I would expect. 

I cannot get to it with the tab key and VoiceOver doesn't read it. I also can't find in the VO landmark menu. I don't know if that part is important or not.

pookmish:
@jenbreese You should be able to tab to it at the very end of the page. Screen reader was able to pick it up for me in chrome just fine.

I'm not sure what you mean by the VO landmark menu.

jenbreese:
@pookmish I gave up too soon because I was on a really long page for tabbing. It worked. 

I meant the Voice Over landmarks menu which I think is a nice way to switch to different landmarks on the page. You have to be using VoiceOver and you press a key combo that brings up a menu of navigation shortcuts on a page. I was just trying all the things. :) 
image

pookmish:
Ok i think if a user's using the landmarks menu, they don't really need the "back to top" since they can just navigate to the top via that. :smile:

jenbreese:
@pookmish I've been trying to get a screenshot so you can see the menu but the screen grab key commands are interfered with VoiceOver.

jenbreese (approved):
Looks and works good for me.

imonroe:
The code seems fine, but what happened with the package-lock.json file? It went from version 2 to version 1, and lost 16,000 lines.

PR #261, opened 2022-02-11
preg_replace of null is deprecated in php 8, use strings

pookmish:
# READY FOR REVIEW

# Summary
- Make sure `preg_replace` is acting on a string instead of `null`

# Need Review By (Date)
- 

# Urgency
- 

# Steps to Test

# See Also

### DIFF
--- a/stanford_basic.theme
+++ b/stanford_basic.theme
@@ -138,8 +138,8 @@ function stanford_basic_preprocess_views_view(&$variables) {
  * @return string
  *   Cleaned string.
  */
-function _stanford_basic_change_characters($string) {
-  return preg_replace("/[^a-zA-Z\d\s]/", '-', $string);
+function _stanford_basic_change_characters($string = '') {
+  return preg_replace("/[^a-zA-Z\d\s]/", '-', $string ?? '');
 }
 
 /**


stanford_basic.theme
```diff
@@ -138,8 +138,8 @@ function stanford_basic_preprocess_views_view(&$variables) {
  * @return string
  *   Cleaned string.
  */
-function _stanford_basic_change_characters($string) {
-  return preg_replace("/[^a-zA-Z\d\s]/", '-', $string);
+function _stanford_basic_change_characters($string = '') {
```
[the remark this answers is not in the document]
sherakama: We should probably have a fallback check on the hook_block call of this function to avoid the null.

stanford_basic.theme
```diff
@@ -138,8 +138,8 @@ function stanford_basic_preprocess_views_view(&$variables) {
...
+  return preg_replace("/[^a-zA-Z\d\s]/", '-', $string ?? '');
```
imonroe, line 142: Couldn't we fix it with: 
```suggestion
  return preg_replace("/[^a-zA-Z\d\s]/", '-', (string) $string ?? '');
```
And leave the method signature alone?

pookmish: Yeah, i suppose that would work. Casting `null` produces an empty string. @sherakama, any objections?

pookmish: I thought about your suggestion @imonroe but technically `$string` could be an array. `preg_replace` accepts arrays for the first three parameters. Although we aren't currently using it that way, i'd rather keep it as a more utility function than a specific one.

pookmish:
I'm just going to close this for now. We can revisit if necessary in the future. "Deprecated" != "Error" right now.

